PR #1, opened 2021-06-18
WIP: Implement Dns record filtering

LaakkonenJussi:
A work-in-progress version of DNS record filtering. ipconfig notifications are sometimes lacking either of IP families so this does need some work still.

This addresses the issue of leaking data to IPv6 network when IPv4 VPN is connected using transport that supports both IP networks. This can happen because the DNS can return IPv6 (AAAA) records as well, and AAAA records are also requested if IPv6 is enabled on an online service. Since IPv6 is preferred over IPv4 the AAAA record is used and connection is made over IPv6 network, bypassing the default service in such case. By filtering out the DNS records based on the networks that the default service has in connected state this issue is prevented from happening.

An option is added to configuration to toggle whether the feature implemented here is enabled or not. By default the option FilterDNSRecords is on (true). TODO: this could be the other way around now as there are other means to prevent using IPv6 with VPNs.

### DIFF
--- a/connman/doc/connman.conf.5.in
+++ b/connman/doc/connman.conf.5.in
@@ -136,6 +136,15 @@ If /sys/class/net/<interface>/uevent does not contain DEVTYPE information,
 heuristics are used to choose between wifi and ethernet device types. If
 neither is appropriate, this setting can be used to provide more suitable
 fallback value - e.g. rndis0:gadget.
+.TP
+.BI FilterDNSRecords= true\ \fR|\fB\ false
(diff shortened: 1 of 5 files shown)


LaakkonenJussi:
Kind of forgot what is to be done with this one. It simply may be that in every situation this does not work and it heavily enforces the current default service's enabled ipconfig types on the filtering. But eventually, that is the effect of this feature..

Need to revisit this later on, and to have it as optional at least.

PR #2, opened 2021-06-23
[service] Do not start autoconnect round without a back-off timer. JB#52947

d-grigorev:
The timer was made to trigger autoconnect round for services that are
in failed state. However, the cellular service does not stay in the failed
state for long and goes idle right after the timer is started due to
gofono's "activate-failed" signal being triggered.

This PR includes two commits: the first makes autoconnect round to start
regardless the service is in idle state, and the second does not allow auto
connection bypassing the back off timer.

### DIFF
--- a/connman/src/service.c
+++ b/connman/src/service.c
@@ -5648,8 +5648,8 @@ static gboolean service_retry_connect(gpointer data)
 	DBG("service %p", service);
 	service->connect_retry_timer = 0;
 
-	if (service->state == CONNMAN_SERVICE_STATE_FAILURE) {
-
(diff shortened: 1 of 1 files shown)


d-grigorev:
The PR is moved here from https://git.sailfishos.org/mer-core/connman/merge_requests/308.

LaakkonenJussi (approved):
This did not seem to break anything in my tests and this new version now looks better. Since this relies on our own retry-logic implemented to `service.c` there is no need to think upstreaming.

Tested with WiFi and VPNs but as earlier discussed, mobile data is quite difficult to get in that shape that the issue described would happen.

LGTM.

PR #3, opened 2021-07-15
[service] Fix missing properties for saved wlans. JB#55059

ikamaletdinov:
If wlan service was saved and restored by connman we can observe
missing wifi props (BSSID, MaxRate, Frequency, EncryptionMode)
after that wlan network was appeared by scan in libconnman-qt.

To fix that explicitly add notifications for wifi props after
adding `network` to saved `service`.

### DIFF
--- a/connman/src/service.c
+++ b/connman/src/service.c
@@ -4,7 +4,7 @@
  *
  *  Copyright (C) 2007-2014  Intel Corporation. All rights reserved.
  *  Copyright (C) 2014-2020  Jolla Ltd.
- *  Copyright (C) 2020  Open Mobile Platform LLC.
+ *  Copyright (C) 2020-2021  Open Mobile Platform LLC.
  *
  *  This program is free software; you can redistribute it and/or modify
  *  it under the terms of the GNU General Public License version 2 as
@@ -3433,6 +3433,27 @@ static void append_properties(DBusMessageIter *dict, dbus_bool_t limited,
 				GET_DEFAULT_ACCESS_ACCESS);
 }
 
+static void append_properties_updated(DBusMessageIter *dict,
+					struct connman_service *service)
+{
+	switch (service->type) {
+	case CONNMAN_SERVICE_TYPE_UNKNOWN:
+	case CONNMAN_SERVICE_TYPE_SYSTEM:
+	case CONNMAN_SERVICE_TYPE_GPS:
+	case CONNMAN_SERVICE_TYPE_VPN:
+	case CONNMAN_SERVICE_TYPE_P2P:
+	case CONNMAN_SERVICE_TYPE_CELLULAR:
+	case CONNMAN_SERVICE_TYPE_ETHERNET:
+	case CONNMAN_SERVICE_TYPE_BLUETOOTH:
+	case CONNMAN_SERVICE_TYPE_GADGET:
+		break;
+	case CONNMAN_SERVICE_TYPE_WIFI:
+		if (service->network)
+			append_wifi_ext_info(dict, service->network);
+		break;
+	}
+}
+
 static void append_struct_service(DBusMessageIter *iter,
 		connman_dbus_append_cb_t function,
 		struct connman_service *service)
@@ -3468,6 +3489,14 @@ static void append_struct(gpointer value, gpointer user_data)
 		return;
 
 	append_struct_service(iter, append_dict_properties, service);
+}
+
+static void append_dict_properties_updated(DBusMessageIter *dict,
+							void *user_data)
+{
+	struct connman_service *service = user_data;
+
+	append_properties_updated(dict, service);
 }
 
 void __connman_service_list_struct(DBusMessageIter *iter)
@@ -5978,6 +6007,7 @@ static struct _services_notify {
 	int id;
 	GHashTable *add;
 	GHashTable *remove;
+	GHashTable *update;
 } *services_notify;
 
 
@@ -5996,6 +6026,11 @@ static void service_append_added_foreach(gpointer data, gpointer user_data)
 
 		append_struct(service, iter);
 		g_hash_table_remove(services_notify->add, service->path);
+	} else if (g_hash_table_lookup(services_notify->update, service->path)) {
+		DBG("updated %s", service->path);
+
+		append_struct_service(iter, append_dict_properties_updated, service);
+		g_hash_table_remove(services_notify->update, service->path);
 	} else {
 		DBG("changed %s", service->path);
 
@@ -6045,6 +6080,7 @@ static gboolean service_send_changed(gpointer data)
 
 	g_hash_table_remove_all(services_notify->remove);
 	g_hash_table_remove_all(services_notify->add);
+	g_hash_table_remove_all(services_notify->update);
 
 	return FALSE;
 }
@@ -6279,6 +6315,7 @@ static void service_schedule_added(struct connman_service *service)
 	DBG("service %p", service);
 
 	g_hash_table_remove(services_notify->remove, service->path);
+	g_hash_table_remove(services_notify->update, service->path);
 	g_hash_table_replace(services_notify->add, service->path, service);
 
 	service_schedule_changed();
@@ -6294,8 +6331,25 @@ static void service_schedule_removed(struct connman_service *service)
 	DBG("service %p %s", service, service->path);
 
 	g_hash_table_remove(services_notify->add, service->path);
+	g_hash_table_remove(services_notify->update, service->path);
 	g_hash_table_replace(services_notify->remove, g_strdup(service->path),
 			NULL);
+
+	service_schedule_changed();
+}
+
+static void service_schedule_updated(struct connman_service *service)
+{
+	/* Only update if service has path */
+	if (!service || !service->path) {
+		return;
+	}
+
+	DBG("service %p", service);
+
+	g_hash_table_remove(services_notify->remove, service->path);
+	g_hash_table_remove(services_notify->add, service->path);
+	g_hash_table_replace(services_notify->update, service->path, service);
 
 	service_schedule_changed();
 }
@@ -9241,6 +9295,7 @@ static void update_from_network(struct connman_service *service,
 		service->network = connman_network_ref(network);
 		connman_network_autoconnect_changed(service->network,
 							service->autoconnect);
+		service_schedule_updated(service);
         }
 
 	if (was_available != service_available.value(service))
@@ -9731,6 +9786,7 @@ int __connman_service_init(void)
 	services_notify->remove = g_hash_table_new_full(g_str_hash,
 			g_str_equal, g_free, NULL);
 	services_notify->add = g_hash_table_new(g_str_hash, g_str_equal);
+	services_notify->update = g_hash_table_new(g_str_hash, g_str_equal);
 
 	remove_unprovisioned_services();
 
@@ -9780,6 +9836,7 @@ void __connman_service_cleanup(void)
 
 	g_hash_table_destroy(services_notify->remove);
 	g_hash_table_destroy(services_notify->add);
+	g_hash_table_destroy(services_notify->update);
 	g_free(services_notify);
 
 	dbus_connection_unref(connection);


LaakkonenJussi:
Is there a task id for this? It is missing from the title/commit.

connman/src/service.c
```diff
@@ -9244,6 +9244,10 @@ static void update_from_network(struct connman_service *service,
 		service->network = connman_network_ref(network);
 		connman_network_autoconnect_changed(service->network,
 							service->autoconnect);
+		__connman_service_network_property_changed(service, "BSSID");
+		__connman_service_network_property_changed(service, "MaxRate");
+		__connman_service_network_property_changed(service, "Frequency");
+		__connman_service_network_property_changed(service, "EncryptionMode");
```
LaakkonenJussi, line 9250: I don't think this is the right place for these. These are appended to the DBus struct by `append_wifi_ext_info()` and the struct is then sent when service.c schedules changes about the services to be sent. Those calls to `__connman_service_network_property_changed()` would in this case just send a value update even when the value has not been changed.

If the intention is to get these values updated when the new network for a WiFi service is established I'm pretty sure that this is covered by the scheduled property changes of the services. network.c does send the updates when the values change, at least for `BBSID` and `Frequency` as they can change during the established connection to the network.

In what scenario this issue reproduces and why these are needed? As there is no task to refer to I'm kind of shooting half blindly into the dark with my comments.

ikamaletdinov:
> Is there a task id for this? It is missing from the title/commit.

Hello, I have created bug report for this BUG: JB#55059

connman/src/service.c
ikamaletdinov: Your are right, these properties are appended but under normal conditions. Saved wifi `services` are added to connman services list after device start. And after `saved` network adds to already existed `service` we have missing BSSID and other props in mw layer (libconnman-qt, mdm and etc.). Thats why I suggested to explicitly send notifications for these properties when we replace `network` (that was NULL) in existing `saved` `service`. Otherwise these properties always empty (NULL) in mw layer.

LaakkonenJussi: Since this is the case, thanks for the explanation, I see that this issue happens because the service is not a new, added service, for which all the properties are added by calling [`append_struct()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L3261)b in [`services_append_added_foreach()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L5984). This happens because [`append_struct_service()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L3436) without any callback just adds the path of an existing service, this is called by [`append_struct()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L3261) with `append_dict_properties()` as callback so if the service is in added it will get have the properties updated. The final call here is to `service_list_sort()` which in case of more than one service triggers the sending of changed.

However, this is originally triggered via network.c by calling `__connman_service_create_network()` and if there is no `path` defined the service is treated as a new one, but if the service has path, so the [network comes back](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L9291) it should every time trigger the `service_schedule_added()` as it is done for the services without `path` set. However, documentation ([manager-api.txt](https://github.com/sailfishos/connman/blob/master/connman/doc/manager-api.txt)) does say that the dict for a service may be empty or contain only changed values. In case of a returning network, these property changed values you have here are those that should be informed. However, this would first send the property changed values and only after the service list, which is done in wrong order and I think these should be included into the main services changed list as properties.

This just gave me an idea of using an `updated` hash table value to `struct services_notify` similar to the `added` with an exception that in [`services_append_added_foreach()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L5984) we'd call another function that simply adds a different callback to the [`append_struct_service()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L3436) to include certain updated values based on the service type. And here with WiFi services it could call the [`append_wifi_ext_info()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L2994). However, [`allow_property_changed()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L6310) suggests using add notify for updates. 

That idea would be then extendable to other service types which may need some additional information added as changed values to the `ServicesChanged` signal by just simply amending the `switch()-case` made for the services that need an update on small range of variables. Any comments on this approach?

LaakkonenJussi: And it would require creating a new `service_schedule_updated()` which is then called in [`__connman_service_create_from_network()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L9298) when the service has a path set and needs only part of the property changes to be done.

And like with adding a new service to the `ServicesChanged` signal in [`service_append_added_foreach()`](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L5994) the `updated` could be added by simply calling that `append_struct_service(iter, append_updated_properties, service` where `append_updated_properties()` is the new function to call a specific dict amending function to add the commonly changed parts for the service based on its type. For WiFi, just call `append_wifi_ext_info()`. 

I hope this clarifies what I had in mind.

The fully optimal situation would be that each variable for each service would have its own struct with an additional field `bool updated` which could be used to check if a certain parameter for a service has been changed since the services were last updated to clients listening over D-Bus. But that is beyond the scope here :)

connman/src/service.c
```diff
@@ -5996,6 +6026,10 @@ static void service_append_added_foreach(gpointer data, gpointer user_data)
 
 		append_struct(service, iter);
 		g_hash_table_remove(services_notify->add, service->path);
+	} else if (g_hash_table_lookup(services_notify->update, service->path)) {
+		DBG("updated %s", service->path);
+
+		append_struct_service(iter, append_dict_properties_updated, service);
```
LaakkonenJussi, line 6032: I guess here this could follow the same approach as adding, so `g_hash_table_remove(services_notify->update, service->path);` could be here for consistency?

connman/src/service.c
```diff
@@ -3433,6 +3433,27 @@ static void append_properties(DBusMessageIter *dict, dbus_bool_t limited,
 				GET_DEFAULT_ACCESS_ACCESS);
 }
 
+static void append_properties_updated(DBusMessageIter *dict,
+	struct connman_service *service)
```
LaakkonenJussi, line 3437: [Coding style M4](https://github.com/sailfishos/connman/blob/master/connman/doc/coding-style.txt#L78) instructs to intend these new lines to the right.

connman/src/service.c
```diff
@@ -3470,6 +3491,14 @@ static void append_struct(gpointer value, gpointer user_data)
 	append_struct_service(iter, append_dict_properties, service);
 }
 
+static void append_dict_properties_updated(DBusMessageIter *dict,
+	void *user_data)
```
LaakkonenJussi, line 3495: [Coding style M4](https://github.com/sailfishos/connman/blob/master/connman/doc/coding-style.txt#L78) instructs to intend these new lines to the right.

connman/src/service.c
```diff
@@ -3433,6 +3433,27 @@ static void append_properties(DBusMessageIter *dict, dbus_bool_t limited,
...
+	case CONNMAN_SERVICE_TYPE_GPS:
+	case CONNMAN_SERVICE_TYPE_VPN:
+	case CONNMAN_SERVICE_TYPE_P2P:
+	case CONNMAN_SERVICE_TYPE_CELLULAR:
+	case CONNMAN_SERVICE_TYPE_ETHERNET:
+	case CONNMAN_SERVICE_TYPE_BLUETOOTH:
+	case CONNMAN_SERVICE_TYPE_GADGET:
+		break;
+	case CONNMAN_SERVICE_TYPE_WIFI:
+		if (service->network != NULL)
```
LaakkonenJussi, line 3451: For simplicity this check can be just `if (service->network)`

LaakkonenJussi: Which is also suggested by [coding style M13](https://github.com/sailfishos/connman/blob/master/connman/doc/coding-style.txt#L247).

connman/src/service.c
LaakkonenJussi: Hi, great update. Looking good and I added only some small notes there. mostly about the coding style. It might be even that this could be attempted to be upstreamed, I need to check if there are some contradicting changes or does it suffer from the same symptoms but I guess this would bring value also there.

But first, lets get this completed.

LaakkonenJussi:
One thing that came to mind that should an update on the values prevent the property changed from being sent? I'm referring to: [allow_property_changed()](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L6303) - so when there is an update to an value it shouldn't be possible to send the property signal.

It also dawned to me that the `allow_property_changed()` would have otherwise prevented the sending of the property changed signals as it was your original idea but in that case the service was not added to the `ServicesChanged` signal and that is why it worked then. :)

Anyways, I'd say this is good for further use and brings some clear extendability for the services that may need their properties updated.

ikamaletdinov:
> One thing that came to mind that should an update on the values prevent the property changed from being sent? I'm referring to: [allow_property_changed()](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L6303) - so when there is an update to an value it shouldn't be possible to send the property signal.
> 
> It also dawned to me that the `allow_property_changed()` would have otherwise prevented the sending of the property changed signals as it was your original idea but in that case the service was not added to the `ServicesChanged` signal and that is why it worked then. :)
> 
> Anyways, I'd say this is good for further use and brings some clear extendability for the services that may need their properties updated.

I'm still not sure that it is optimal solution. Also I previosly mentioned that for some time we will have `available` wlan network with empty bssid until ServicesChanged arrived. (wlan becomes `available` (prop update) -> ~100ms -> bssid arrives with ServicesChanged). To mitigate that we will need to block property changes until `ServicesChanged` arrives. So if we block sending properties with allow_property_changed() we will need to add some props to be sent with ServicesChanged signal. I suggest to add `available` , `autoconnect` and `signal strenght` at least. If we miss some props update on that period of time I'm sure that we will get much more unpredictable bugs.

Thats why my first solution with simply 3 props update is not looking so bad :)

ikamaletdinov:
> One thing that came to mind that should an update on the values prevent the property changed from being sent? I'm referring to: [allow_property_changed()](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L6303) - so when there is an update to an value it shouldn't be possible to send the property signal.
> 
> It also dawned to me that the `allow_property_changed()` would have otherwise prevented the sending of the property changed signals as it was your original idea but in that case the service was not added to the `ServicesChanged` signal and that is why it worked then. :)
> 
> Anyways, I'd say this is good for further use and brings some clear extendability for the services that may need their properties updated.

So I suggest to: 
1) Do not block props update just not to add new bugs :) (can be refactored later)
2) Revert back to my first solution :)
3) Send some kind of full state for that network like with `add` procedure. But is makes current `update` algo useless.

connman/src/service.c
ikamaletdinov: Thanks a lot for thre review, I have fixed that test that, everything works fine

LaakkonenJussi:
> > One thing that came to mind that should an update on the values prevent the property changed from being sent? I'm referring to: [allow_property_changed()](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L6303) - so when there is an update to an value it shouldn't be possible to send the property signal.
> > It also dawned to me that the `allow_property_changed()` would have otherwise prevented the sending of the property changed signals as it was your original idea but in that case the service was not added to the `ServicesChanged` signal and that is why it worked then. :)
> > Anyways, I'd say this is good for further use and brings some clear extendability for the services that may need their properties updated.
> 
> I'm still not sure that it is optimal solution. Also I previosly mentioned that for some time we will have `available` wlan network with empty bssid until ServicesChanged arrived. (wlan becomes `available` (prop update) -> ~100ms -> bssid arrives with ServicesChanged). To mitigate that we will need to block property changes until `ServicesChanged` arrives. So if we block sending properties with allow_property_changed() we will need to add some props to be sent with ServicesChanged signal. I suggest to add `available` , `autoconnect` and `signal strenght` at least. If we miss some props update on that period of time I'm sure that we will get much more unpredictable bugs.
> 
> Thats why my first solution with simply 3 props update is not looking so bad :)
> 
> So I suggest to:
> 
> 1. Do not block props update just not to add new bugs :) (can be refactored later)
> 2. Revert back to my first solution :)
> 3. Send some kind of full state for that network like with `add` procedure. But is makes current `update` algo useless.

Let me it put it this way: this new approach we have here is following the way how ConnMan's service changed should operate, at least the documentation supports suggests this but before there was no method to update only service type specific variables. By allowing to send property changes separately when the network is returned might create a situation where there is no previous information about the service itself on some mw component and thus, I would not go to the original idea. It may be that a new middleware component comes up during that period that it either hasn't processed/received `ServicesChanged` signal and now it gets a `PropertyChanged` signal from service that it does not yet know. I guess in such situation the mw component should ignore the `PropertyChanged` signals if it does yet know the service for which the change is for.

I keep scratching my head with this preventing or not preventing property changes. If all property changes on a service are prevented when only a subset of those are being updated then that would create some issues and bugs as you said. So that might be actually wrong - good thing that we started this discussion on the matter. If the `ServicesChanged` comes later with the updated values than the `PropertyChanged` they do have the same values as the `ServicesChanged` is constructed using the same data so either way the mw component listening to the changes would have the same data. If it does the processing in a way that `PropertyChanged` from a yet unknown service is ignored the `ServicesChanged` signal has the same data and no variation happens there. Similarly, if the mw component allows updates to unknown services with  the `PropertyChanged` then the `ServicesChanged` signal again brings no difference as both contain the same value. 

So as a summary: this current one looks to be a fine solution for the issue. Let me test it a bit first.

I think we should not worry about the `100ms` delay at all as it is the normal delay for service updates in ConnMan. By following the original approaches that seemed to work we are not likely to break anything, right? :)

And about the upstreaming, my bad, I wasn't aware that those additional WiFi parameters have not been implemented to upstream `service.c`/wifi plugin. I guess the reason is that they disband `wpa_supplicant` in favor of `iwd`. So we should forget that one, unfortunately.

connman/src/service.c
ikamaletdinov: Thanks a lot for the review, I have fixed that

connman/src/service.c
ikamaletdinov: Thanks a lot for the review, I have fixed that

connman/src/service.c
ikamaletdinov: Nice catch, thanks, I have fixed that

ikamaletdinov:
> Let me it put it this way: this new approach we have here is following the way how ConnMan's service changed should operate, at least the documentation supports suggests this but before there was no method to update only service type specific variables. By allowing to send property changes separately when the network is returned might create a situation where there is no previous information about the service itself on some mw component and thus, I would not go to the original idea. It may be that a new middleware component comes up during that period that it either hasn't processed/received `ServicesChanged` signal and now it gets a `PropertyChanged` signal from service that it does not yet know. I guess in such situation the mw component should ignore the `PropertyChanged` signals if it does yet know the service for which the change is for.
> 
> I keep scratching my head with this preventing or not preventing property changes. If all property changes on a service are prevented when only a subset of those are being updated then that would create some issues and bugs as you said. So that might be actually wrong - good thing that we started this discussion on the matter. If the `ServicesChanged` comes later with the updated values than the `PropertyChanged` they do have the same values as the `ServicesChanged` is constructed using the same data so either way the mw component listening to the changes would have the same data. If it does the processing in a way that `PropertyChanged` from a yet unknown service is ignored the `ServicesChanged` signal has the same data and no variation happens there. Similarly, if the mw component allows updates to unknown services with the `PropertyChanged` then the `ServicesChanged` signal again brings no difference as both contain the same value.
> 
> So as a summary: this current one looks to be a fine solution for the issue. Let me test it a bit first.
> 
> I think we should not worry about the `100ms` delay at all as it is the normal delay for service updates in ConnMan. By following the original approaches that seemed to work we are not likely to break anything, right? :)
> 
> And about the upstreaming, my bad, I wasn't aware that those additional WiFi parameters have not been implemented to upstream `service.c`/wifi plugin. I guess the reason is that they disband `wpa_supplicant` in favor of `iwd`. So we should forget that one, unfortunately.

Thanks a lot for the explanation. I have tested that solution and it is working fine and mitigates that bug. MW now can see BSSID as expected. You can check it with d-bus monitor also to see some lag between becoming service `available` and `ServicesChanged` signal. And It doesn't break anything or course. (previously there was empty BSSID in that case)
This is nice point for future upgrade to mark properties as `dirty` in case of change before `ServicesChanged` signal.

connman/src/service.c
LaakkonenJussi: Could you intend this to the right up to the 80 char limit in style of, e.g. [append_ext_wifi_info()](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L2994) ?

I know the code has in itself different styles but each new functionality we add should at least try to follow the current coding styles. My point here is that at some point we'd like to upstream as much as possible we have and when we do the style changes at the steps of adding new code we probably avoid one additional review round in the future :)

connman/src/service.c
LaakkonenJussi: Similar comment here as in https://github.com/sailfishos/connman/pull/3#discussion_r688444799

LaakkonenJussi:
> 
> Thanks a lot for the explanation. I have tested that solution and it is working fine and mitigates that bug. MW now can see BSSID as expected. You can check it with d-bus monitor also to see some lag between becoming service `available` and `ServicesChanged` signal. And It doesn't break anything or course. (previously there was empty BSSID in that case)
> This is nice point for future upgrade to mark properties as `dirty` in case of change before `ServicesChanged` signal.

You are welcome. I tested this and seems to work well as you said. New properties appear in the message.

LGTM, beside the small formatting nitpick. Once you do those minor ones I'll merge this to master. Good to get these kind of things fixed.

connman/src/service.c
ikamaletdinov: Thanks a lot. I have created tabs like in surrounded code, but it seems that it was created so long time ago :)

connman/src/service.c
ikamaletdinov: Thanks a lot. I have fixed that.

LaakkonenJussi: Great!

connman/src/service.c
LaakkonenJussi: Looking good now.

LaakkonenJussi (commented):
LGTM.

PR #4, opened 2021-08-03
Handle ENODATA in systemd login monitor.

LaakkonenJussi:
This change adds handling of `ENODATA` as an error that does not result in quitting the periodic process of checking active user. This may happen when systemd has not yet got the complete data (e.g., `ACTIVE_UID` is missing from the seat).


LaakkonenJussi:
@inzanity review appreciated.

spiiroin (approved):

PR #5, opened 2021-08-06
Improve invalid auth and disconnect notify, fix cb use in PPPD based VPN plugins

LaakkonenJussi:
[l2tp] Improve invalid auth and disconnect notify, fix cb use. JB#52482
[pptp] Improve invalid auth and disconnect notify, fix cb use. JB#52482
[openfortivpn] Improve invalid auth and disconnect notify, fix cb use. JB#52482

All follow the same codebase, same fixes apply to all. This fixes issue of VPN not dying in case the auth is invalid.

These changes enforce quitting of the VPN plugin when it is being disconnected, also because of invalid authentication and then prompts for new password via VPN agent. These fix premature calling of the connect callback before connection status is being evaluated at all. Other minor cleanups and fixes here and there.

Slight increase in memory consumption because of storing the private variables for the duration of the connection. It seems that these three plugins would greatly benefit from a few generalized helper/util/lib functions.

### DIFF
--- a/connman/vpn/plugins/l2tp.c
+++ b/connman/vpn/plugins/l2tp.c
@@ -4,7 +4,7 @@
  *
  *  Copyright (C) 2010,2013  BMW Car IT GmbH.
  *  Copyright (C) 2012-2013  Intel Corporation. All rights reserved.
- *  Copyright (C) 2019  Jolla Ltd.
+ *  Copyright (C) 2019-2021  Jolla Ltd.
  *  Copyright (C) 2019  Open Mobile Platform LLC.
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -125,11 +125,39 @@ struct {
 static DBusConnection *connection;
 
 struct l2tp_private_data {
+	struct vpn_provider *provider;
 	struct connman_task *task;
 	char *if_name;
 	vpn_provider_connect_cb_t cb;
 	void *user_data;
 };
+
+static void l2tp_connect_done(struct l2tp_private_data *data, int err)
+{
+	vpn_provider_connect_cb_t cb;
+	void *user_data;
+
+	if (!data || !data->cb)
+		return;
+
+	/* Ensure that callback is called only once */
+	cb = data->cb;
+	user_data = data->user_data;
+	data->cb = NULL;
+	data->user_data = NULL;
+	cb(data->provider, user_data, err);
+}
+
+static void free_private_data(struct l2tp_private_data *data)
+{
+	if (vpn_provider_get_plugin_data(data->provider) == data)
+		vpn_provider_set_plugin_data(data->provider, NULL);
+
+	l2tp_connect_done(data, EIO);
+	vpn_provider_unref(data->provider);
+	g_free(data->if_name);
+	g_free(data);
+}
 
 static DBusMessage *l2tp_get_sec(struct connman_task *task,
 			DBusMessage *msg, void *user_data)
@@ -168,6 +196,9 @@ static int l2tp_notify(DBusMessage *msg, struct vpn_provider *provider)
 	char *addressv4 = NULL, *netmask = NULL, *gateway = NULL;
 	char *ifname = NULL, *nameservers = NULL;
 	struct connman_ipaddress *ipaddress = NULL;
+	struct l2tp_private_data *data;
+
+	data = vpn_provider_get_plugin_data(provider);
 
 	dbus_message_iter_init(msg, &iter);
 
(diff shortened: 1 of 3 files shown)


LaakkonenJussi:
@inzanity review appreciated. Also @okodron / @a-andreyev review + testing help appreciated, since I have no working Fortinet VPN at my disposal. Basically all these three work in the same way as they use pppd but it does not hurt to get a verification that this change works well with openfortivpn either.

connman/vpn/plugins/l2tp.c
```diff
@@ -762,20 +801,29 @@ static void request_input_cb(struct vpn_provider *provider,
...
+	data = g_try_new0(struct l2tp_private_data, 1);
+	if (!data)
+		return -ENOMEM;
+
+	vpn_provider_set_plugin_data(provider, data);
+	data->provider = vpn_provider_ref(provider);
+	data->task = task;
+	data->if_name = g_strdup(if_name);
+	data->cb = cb;
+	data->user_data = user_data;
```
inzanity, line 825: To me it would sound more logical to first populate the data and set after, but as this is all single threaded stuff, it don't make much of a difference.

connman/vpn/plugins/l2tp.c
```diff
@@ -125,12 +125,40 @@ struct {
...
+	data->user_data = NULL;
+	cb(data->provider, user_data, err);
+}
+
+static void free_private_data(struct l2tp_private_data *data)
+{
+	if (vpn_provider_get_plugin_data(data->provider) == data)
+		vpn_provider_set_plugin_data(data->provider, NULL);
+
+	l2tp_connect_done(data, EIO);
```
inzanity, line 156: I'm not entirely sure whether I like this call, but I guess it's better safe than sorry.

inzanity:
Can't spot anything obviously wrong with it at least. Indeed some refactoring to reduce code duplication would be smart later on.

inzanity (approved):

connman/vpn/plugins/l2tp.c
LaakkonenJussi: This is for the sake of binary simply crashing to make sure that the callback and D-Bus call is replied to. And other odd cases when the VPN dies because of, e.g., invalid content passed to the binary via conf file. If it is already called, then there is nothing to do. But because of varying limits for pending D-Bus calls I think this is almost a must to have here as well. Just to be sure :).

connman/vpn/plugins/l2tp.c
LaakkonenJussi: I copied the approach from some other plugin, worked there but maybe it could be put after all this anyways.

LaakkonenJussi:
> Can't spot anything obviously wrong with it at least. Indeed some refactoring to reduce code duplication would be smart later on.

I think the library use would reduce the code a lot in the plugins. I constantly see issues that are identical with some plugins. But that is work to be done later on.

I'll fix that minor ordering issue in adding the data and propose this to upstream as well.

connman/vpn/plugins/l2tp.c
LaakkonenJussi: I guess, explained? :)

LaakkonenJussi:
Submitted to upstream https://lore.kernel.org/connman/20210813134209.23253-1-jussi.laakkonen@jolla.com/

 @okodron / @a-andreyev any comments on the OpenFortiVPN changes?

okodron:
> Submitted to upstream https://lore.kernel.org/connman/20210813134209.23253-1-jussi.laakkonen@jolla.com/
> 
> @okodron / @a-andreyev any comments on the OpenFortiVPN changes?

It looks ok from the source code point of view. Unfortunately didn't have time to check it yet with running FortiNet.

connman/vpn/plugins/l2tp.c
```diff
@@ -125,12 +125,40 @@ struct {
...
+
+	/* Ensure that callback is called only once */
+	cb = data->cb;
+	user_data = data->user_data;
+	data->cb = NULL;
+	data->user_data = NULL;
+	cb(data->provider, user_data, err);
+}
+
+static void free_private_data(struct l2tp_private_data *data)
```
spiiroin, line 151: I wish also allocations were encapsulated in functions... would not need to go back and forth between here and inlined allocation code to see that what is allocated also gets released. But I guess this is how it is, matter of opinion,  and does not affect correctness of the code.

spiiroin (approved):
Changes make sense, can't spot any obvious problems.

connman/vpn/plugins/l2tp.c
LaakkonenJussi: There is a task where a library/utils for the VPNs is to be created and this is one part of it as these private data structs across every plugin have a lot of similarities and mostly same content. These need a major refactoring then, oh well, upstream approved these l2tp and pptp changes as is.

okodron (approved):

PR #6, opened 2021-08-12
Implement auth error heuristic and per provider value

LaakkonenJussi:
Implement heuristic for auth error counter to avoid losing credentials because of misbehaving server reporting back `AUTH_FAILED` control message. This is achieved by adding "`AuthErrorLimit`" that can be set by user via D-Bus or by a VPN plugin. By default one (1) auth error is allowed until credentials are re-requested (or cleared by the VPN agent). OpenVPN sets this value by default to 10 unless already defined and can be disabled by setting value zero (0). Some VPNs may not require this option and that is why lenience of one such error is allowed by default. This feature is included only when the previous connection with the provider has been successful but there are authentication errors reported as a result of server denying the connection by some specific reason. This does not affect to disconnections etc.

This was noticed to happen in a scenario where an OpenVPN (using ProtonVPN) is connected over UDP (and it seems that with TCP this happens as well) using cellular, connection is switched to WiFi letting VPN to connect, triggering cellular off and back on again, disconnecting WiFi and letting VPN to connect over cellular after which WiFi is re-enabled. Usually this is enough to time the network disconnects in the sense that OpenVPN binary cannot send the disconnect message back to server having following in the system log:
```
    openvpn: event_wait : Interrupted system call (code=4)
    openvpn: SIGTERM received, sending exit notification to peer
    openvpn: write UDP: Network is unreachable (code=101)
    openvpn: Closing TUN/TAP interface
```

As a solution each provider sets a run-time only `previous_connect_time` using the monotonic boot time clock whenever the connection has been successful. This, in conjunction with a limit for allowed authentication errors is used to determine whether the actual authentication error count reported to VPN agent or not. After a connection has been made all errors are cleared. If the connection abruptly goes away, e.g., the server is lost only the `conn_error_counter` is increased and `previous_connect_time` is cleared but this does not trigger clearing of the authentication error and, provided the credentials are still valid, next attempt succeeds whet the server is again reachable. Same applies for a proper disconnect. Thus, this solution does not interfere with normal cases but mitigates unwanted credential loss in case if the VPN server determines connection limit being full and reports back with authentication error message.

### DIFF
--- a/connman/vpn/plugins/openvpn.c
+++ b/connman/vpn/plugins/openvpn.c
@@ -1114,6 +1114,15 @@ static int ov_connect(struct vpn_provider *provider,
 	const char *tmpdir;
 	struct ov_private_data *data;
 
+	/*
+	 * Explicitly set limit of 10 for authentication errors. This defines
+	 * the authentication error message limit from the server before VPN
+	 * agent is instructed to clear the credentials. This is effective only
+	 * after a succesful connection has been made within CONNECT_OK_DIFF
+	 * time. User defined value for "AuthErrorLimit" overrides this.
+	 */
+	vpn_provider_set_auth_error_limit(provider, 10);
+
 	data = g_try_new0(struct ov_private_data, 1);
 	if (!data)
 		return -ENOMEM;
--- a/connman/vpn/vpn-provider.c
+++ b/connman/vpn/vpn-provider.c
@@ -3,7 +3,7 @@
  *  ConnMan VPN daemon
  *
  *  Copyright (C) 2012-2013  Intel Corporation. All rights reserved.
- *  Copyright (C) 2019-2020  Jolla Ltd.
+ *  Copyright (C) 2019-2021  Jolla Ltd.
  *  Copyright (C) 2019  Open Mobile Platform LLC.
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -2091,6 +2091,21 @@ int vpn_provider_indicate_error(struct vpn_provider *provider,
 {
 	DBG("provider %p id %s error %d", provider, provider->identifier,
 									error);
+
+	/*
+	 * Ignore adding of errors when the VPN is idle or not set. Calls may
+	 * happen in a case when networks are rapidly changed and the call to
+	 * vpn_died() is done before executing the connect_cb() from the
+	 * plugin. Then vpn.c:vpn_died() executes the plugin specific died()
+	 * function which may free the plugin private data, containing also
+	 * the callback which hasn't yet been called. As a result the provider
+	 * might already been reset to idle state when the callback is executed
+	 * resulting in unnecessary reset of the previous succesful connect
+	 * timer and adding of an error for already disconnected VPN.
+	 */
+	if (provider->state == VPN_PROVIDER_STATE_IDLE ||
+				provider->state == VPN_PROVIDER_STATE_UNKNOWN)
+		return 0;
 
 	vpn_provider_set_state(provider, VPN_PROVIDER_STATE_FAILURE);
 
--- a/connman/vpn/vpn-provider.c
+++ b/connman/vpn/vpn-provider.c
@@ -25,6 +25,14 @@
 #include <config.h>
 #endif
 
+/*
+ * One hour difference in seconds between connections for checking whether to
+ * treat the authentication error as a real error or as a result of rapid
+ * transport change.
+ */
+#define CONNECT_OK_DIFF			(time_t)60*60
+#define AUTH_ERROR_LIMIT_DEFAULT	1
+
 #define STATE_INTERVAL_DEFAULT		0
 
 #define CONNMAN_STATE_ONLINE 		"online"
@@ -40,6 +48,7 @@
 #include <connman/log.h>
 #include <gweb/gresolv.h>
 #include <netdb.h>
+#include <time.h>
 
 #include "../src/connman.h"
 #include "connman/agent.h"
@@ -101,6 +110,8 @@ struct vpn_provider {
 	unsigned int auth_error_counter;
 	unsigned int conn_error_counter;
 	unsigned int signal_watch;
+	unsigned int auth_error_limit;
+	time_t previous_connect_time;
 };
 
 struct vpn_provider_connect_data {
@@ -1302,12 +1313,15 @@ static void reset_error_counters(struct vpn_provider *provider)
 	if (!provider)
 		return;
 
+	DBG("provider %p", provider);
+
 	provider->auth_error_counter = provider->conn_error_counter = 0;
 }
 
 static int vpn_provider_save(struct vpn_provider *provider)
 {
 	GKeyFile *keyfile;
+	const char *value;
 
 	DBG("provider %p immutable %s", provider,
 					provider->immutable ? "yes" : "no");
@@ -1337,6 +1351,11 @@ static int vpn_provider_save(struct vpn_provider *provider)
 			"Host", provider->host);
 	g_key_file_set_string(keyfile, provider->identifier,
 			"VPN.Domain", provider->domain);
+
+	value = vpn_provider_get_string(provider, "AuthErrorLimit");
+	if (value)
+		g_key_file_set_string(keyfile, provider->identifier,
+						"AuthErrorLimit", value);
 
 	if (provider->user_networks) {
 		gchar **networks;
@@ -1845,6 +1864,18 @@ static void append_dns(DBusMessageIter *iter, void *user_data)
 		append_nameservers(iter, provider->nameservers);
 }
 
+static time_t get_uptime(void)
+{
+	struct timespec t = { 0 };
+
+	if (clock_gettime(CLOCK_BOOTTIME, &t) == -1) {
+		connman_warn("clock_gettime() error %d, uptime failed", errno);
+		return 0;
+	}
+
+	return t.tv_sec;
+}
+
 static int provider_indicate_state(struct vpn_provider *provider,
 				enum vpn_provider_state state)
 {
@@ -1863,6 +1894,8 @@ static int provider_indicate_state(struct vpn_provider *provider,
 	provider->state = state;
 
 	if (state == VPN_PROVIDER_STATE_READY) {
+		provider->previous_connect_time = get_uptime();
+
 		connman_dbus_property_changed_basic(provider->path,
 					VPN_CONNECTION_INTERFACE, "Index",
 					DBUS_TYPE_INT32, &provider->index);
@@ -2075,8 +2108,10 @@ void vpn_provider_add_error(struct vpn_provider *provider,
 {
 	switch (error) {
 	case VPN_PROVIDER_ERROR_UNKNOWN:
+		provider->previous_connect_time = 0;
 		break;
 	case VPN_PROVIDER_ERROR_CONNECT_FAILED:
+		provider->previous_connect_time = 0;
 		++provider->conn_error_counter;
 		break;
 	case VPN_PROVIDER_ERROR_LOGIN_FAILED:
@@ -2084,6 +2119,10 @@ void vpn_provider_add_error(struct vpn_provider *provider,
 		++provider->auth_error_counter;
 		break;
 	}
+
+	DBG("%p connect errors %d auth errors %d", provider,
+						provider->conn_error_counter,
+						provider->auth_error_counter);
 }
 
 int vpn_provider_indicate_error(struct vpn_provider *provider,
@@ -2258,6 +2297,7 @@ static void provider_initialize(struct vpn_provider *provider)
 					g_free, free_route);
 	provider->setting_strings = g_hash_table_new_full(g_str_hash,
 					g_str_equal, g_free, free_setting);
+	provider->auth_error_limit = AUTH_ERROR_LIMIT_DEFAULT;
 }
 
 static struct vpn_provider *vpn_provider_new(void)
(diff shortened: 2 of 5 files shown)


LaakkonenJussi:
@inzanity comments are appreciated. I know that this makes it a bit slower on the VPNs to request new credentials when user disconnects the VPN and server has changed the password. But I think there is no proper way to differentiate between disconnect because of lost transport and disconnect from an user.

An alternative might be to have a callback in `plugins/vpn.c` VPN disconnect reply handling which disconnects the transport about to be disconnected only after VPN has been disconnected. But that case would not work with abruptly disconnected networks. This works in that case.

inzanity:
While this feels wrong, I don't think there's much else we could do.

connman/vpn/plugins/openvpn.c
```diff
@@ -450,6 +450,15 @@ static int run_connect(struct ov_private_data *data,
...
 	int err = 0;
 
+	/*
+	 * Unless set explicitly by the user default to 10 auth errors after a
+	 * succesful connection attempt was made before clearing them via VPN
+	 * agent.
+	 */
+	option = vpn_provider_get_string(provider, "AuthErrorLimit");
+	if (!option)
+		vpn_provider_set_string(provider, "AuthErrorLimit", 10);
```
inzanity, line 460: Should probably be "10".

connman/vpn/vpn-provider.c
```diff
@@ -3289,6 +3301,36 @@ const char *vpn_provider_get_path(struct vpn_provider *provider)
...
+	 * if the underlying network used as transport dies before that and
+	 * there is no route to server anymore. In such case it really seems
+	 * that authentication error is reported back by, at least OpenVPN
+	 * binary but allowing it to try again if we had previously succesful
+	 * connection with the same provider, which will not result in
+	 * potentially unnecessary credential resets.
+	 */
+	if (provider->conn_error_counter == 0 &&
+				provider->previous_connect_status == true &&
+				provider->auth_error_counter < limit) {
```
inzanity, line 3328: I'm guessing this should be <=

connman/vpn/vpn-provider.c
```diff
@@ -3289,6 +3301,36 @@ const char *vpn_provider_get_path(struct vpn_provider *provider)
...
+	 * This crude heuristic is for VPN providers who require that the VPN
+	 * in use does complete clean shutdown process, which is not possible
+	 * if the underlying network used as transport dies before that and
+	 * there is no route to server anymore. In such case it really seems
+	 * that authentication error is reported back by, at least OpenVPN
+	 * binary but allowing it to try again if we had previously succesful
+	 * connection with the same provider, which will not result in
+	 * potentially unnecessary credential resets.
+	 */
+	if (provider->conn_error_counter == 0 &&
```
inzanity, line 3326: If I follow the code correctly, this is redundant and checking just the previous_connect_status would suffice.

LaakkonenJussi: True, I got an idea that what if the `previous_connect` would hold the time of previous connect. Then when the time diff is, lets say, an hour, the first authentication error should be legit and credentials would have to be cleared?

inzanity: Sounds reasonable to me. Maybe use monotonic time just to be sure, as this is a run-time variable it should work and would not break on clock adjustments (quite unlikely case, but as time can be synced from cellular, it is not unthinkable)

connman/vpn/plugins/openvpn.c
LaakkonenJussi: Oops, I guess I missed also the gcc warning as well. Fixed.

LaakkonenJussi:
> While this feels wrong, I don't think there's much else we could do.

Yeah, it is far from ideal solution but unless there would be something in the error codes of OpenVPN to indicate that closing was not completed successfully because of unknown reasons there would then in turn indicate that server might treat us differently from the previous attempt. Need to think if that one hour in between connections would be too much.

connman/vpn/vpn-provider.c
LaakkonenJussi: Removed that and added use of monotonic time with bc93dcd0c1ea3433cf623d7fcec141559eac5fd5

connman/vpn/vpn-provider.c
LaakkonenJussi: Yeah, true. Changed with bc93dcd0c1ea3433cf623d7fcec141559eac5fd5

spiiroin (approved):
fwiw lgtm

connman/vpn/vpn-provider.c
```diff
@@ -3289,6 +3308,36 @@ const char *vpn_provider_get_path(struct vpn_provider *provider)
...
+	unsigned int limit;
+	gint64 diff;
+
+	val = vpn_provider_get_string(provider, "AuthErrorLimit");
+	if (val)
+		limit = g_ascii_strtoull(val, NULL, 10);
+	else
+		limit = ALLOW_AUTH_ERRORS_DEFAULT;
+
+	diff = g_get_monotonic_time() - provider->previous_connect_time;
```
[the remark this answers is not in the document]
LaakkonenJussi: Other option is to use the system time, which can be changed when it is updated via NTP.

Yeah, this may not be really optimal solution yet. Maybe the value just should be checked for being below the `CONNECT_OK_DIFF` to trigger the check as well.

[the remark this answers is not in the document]
LaakkonenJussi: Added a check for using the `auth_error_counter` when the monotonic time is less than the `CONNECT_OK_DIFF` with no previous connection time set: 45e1bdeace1e1ebd61090c78d30947ab0d48d635 - I guess this would suffice for that case you brought up?

connman/vpn/plugins/openvpn.c
```diff
@@ -450,6 +450,15 @@ static int run_connect(struct ov_private_data *data,
...
 	int err = 0;
 
+	/*
+	 * Unless set explicitly by the user default to 10 auth errors after a
+	 * succesful connection attempt was made before clearing them via VPN
+	 * agent.
+	 */
+	option = vpn_provider_get_string(provider, "AuthErrorLimit");
+	if (!option)
+		vpn_provider_set_string(provider, "AuthErrorLimit", "10");
```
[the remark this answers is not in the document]
LaakkonenJussi: In the plugin all of these are hard-coded... Somehow it might be easier in this case to read what is set by default by looking the hard-coded ones?

connman/vpn/vpn-provider.c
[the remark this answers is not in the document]
spiiroin: I guess it might suffice, but ...

It depends on what that glib thingy really does. If it uses CLOCK_MONOTONIC, it does among other things not advance when device is suspended. Which - come to think about it - might actually be bigger problem as the 1h limit can get arbitrary long.

One way to normalize situation could be to: replace that glib function with something that uses CLOCK_BOOTTIME and init / reset that previous_connect_time to -CONNECT_OK_DIFF.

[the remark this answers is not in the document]
LaakkonenJussi: Oh really, that is stopped while in suspend? Great.

I just actually noticed that in some cases OpenVPN may throw a improper shutdown error -> connection error -> reset the connection success time back to 0 and when I get to trigger the AUTH_FAILED reported back by a VPN provider it again clears credentials for no reason. This becomes quite tricky,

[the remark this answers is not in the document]
LaakkonenJussi: So there can be two scenarios:

1) server does not get the proper shutdown cmd from client -> next auths for X period of time are replied with AUTH_FAILED (this scenario is handled now)

2) client binary dies in improper way when disconnecting and after this the server detects the client not being valid for the next auths -> the connection error reset last successful connection timer -> auths are cleared without any valid reason when server again replies with AUTH_FAILED.

I'm starting to think that this AUTH_FAILED is indeed related to servers that allow only one active VPN from an account. And the timeouts, especially with UDP, make this a lengthy process to avoid false positives in active connection detection.  I'm assuming the servers clean the client info with clean shutdown. But that other issue, scenario 2, is because of how the plugin + binary work together.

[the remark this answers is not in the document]
LaakkonenJussi: Thinking out loud here: but then again it is completely wrong to accept connection errors when the VPN is already disconnected and residing in IDLE state as it was the case in 2). That seems to happen when WiFi 1 is connected and then rapidly one changes to WiFi 2.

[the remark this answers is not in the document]
LaakkonenJussi: I'd assume a365748da10f061b82ca6fd52d9e41f870600d82 prevents the 2) from happening now. Getting the problems reproduced takes a long time and effort..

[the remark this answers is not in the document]
LaakkonenJussi: Seemed to work as I finally got the issue to reproduce. But yeah, the ambiguous statement on `g_get_monotonic_time()` about sleep is indeed true. Now getting the uptime from `/proc/uptime` would be an option but.. is it really feasible. In case of a sleeping device it might be that 1h becomes more than 2h with the current approach.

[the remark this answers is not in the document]
LaakkonenJussi: @spiiroin Replaced the monotonic clock with call to sysinfo() 5edf3dfd999be2975fe1fb35153ff7ca6d9433a9 - so now it deals in uptime. That should work now well.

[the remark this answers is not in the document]
spiiroin: > sysinfo() ... That should work

probably does, just wondering why you chose such a big gun (gets all kinds of things in addition to uptime) instead of clock_gettime(CLOCK_BOOTTIME)?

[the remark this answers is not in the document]
LaakkonenJussi: Because Harry Callahan also liked to wear a big gun!

Yeah, that `clock_gettime()` is even better, neater.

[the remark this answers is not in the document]
LaakkonenJussi: Changed and updated the whole thing with documentation and all. Any more concerns/comments?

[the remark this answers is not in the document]
spiiroin: > Any more concerns/comments?

nope

LaakkonenJussi:
@inzanity any concerns or shall I merge this?

connman/vpn/vpn-provider.c
```diff
@@ -3286,9 +3339,67 @@ const char *vpn_provider_get_path(struct vpn_provider *provider)
...
+	if (val)
+		limit = g_ascii_strtoull(val, NULL, 10);
+	else
+		limit = ALLOW_AUTH_ERRORS_DEFAULT;
+
+	if (!limit)
+		return false;
+
+	uptime = get_uptime();
+	if (uptime <= CONNECT_OK_DIFF && !provider->previous_connect_time) {
```
inzanity, line 3378: Why the part before &&? Wouldn't this log message make more sense always when previous_connect_time was 0?

connman/vpn/vpn-provider.c
```diff
@@ -3286,9 +3339,67 @@ const char *vpn_provider_get_path(struct vpn_provider *provider)
...
+		DBG("%p errors %d uptime %ld no known prev conn", provider,
+					provider->auth_error_counter, uptime);
+		return false;
+	}
+
+	diff = uptime - provider->previous_connect_time;
+
+	DBG("%p errors %d limit %d time diff %ld", provider,
+				provider->auth_error_counter, limit, diff);
```
inzanity, line 3387: Nit: limit is unsigned; finer-comb nit: I think the only really portable way to printf time_t is: %jd and cast to intmax_t.

connman/vpn/vpn-provider.c
```diff
@@ -1302,12 +1312,15 @@ static void reset_error_counters(struct vpn_provider *provider)
...
 
+	DBG("provider %p", provider);
+
 	provider->auth_error_counter = provider->conn_error_counter = 0;
 }
 
 static int vpn_provider_save(struct vpn_provider *provider)
 {
 	GKeyFile *keyfile;
+	const char* value;
```
inzanity, line 1323: s/* / */

connman/vpn/plugins/openvpn.c
inzanity, line 460: I haven't looked if there's an easy way to do this in any other way, but done this way, the default value ends up being saved in the configuration file, and if we change the default, it won't change for existing services even if the value has not been explicitly set.

inzanity (approved):
LGTM, just a few nits, feel free to ignore (expect the trivial white-space fix)

connman/vpn/vpn-provider.c
LaakkonenJussi: I think this was from some old thingy ... I guess this is fixed with 3c87042b46fd46316de0631d898c0b0c57bf8c8e

connman/vpn/vpn-provider.c
LaakkonenJussi: Thanks, fixed in 3c87042b46fd46316de0631d898c0b0c57bf8c8e

connman/vpn/vpn-provider.c
LaakkonenJussi: Oops, fixed in 3c87042b46fd46316de0631d898c0b0c57bf8c8e

connman/vpn/plugins/openvpn.c
LaakkonenJussi: There is, in 3c87042b46fd46316de0631d898c0b0c57bf8c8e I added `vpn_provider_set_auth_error_limit()` that is for plugins to set their defaults which is used only when "AuthErrorLimit" is not set by user and defaults to `AUTH_ERROR_LIMIT_DEFAULT` (renamed this as well) that is now 1. Added use of this to OpenVPN as well and to my real surprise, that AUTH_FAILED does occur also with ProtonVPN over TCP! Odd. I was under the assumption that this happens only with UDP hence the protocol design but it could be easily detected when TCP is used. Oh well. Maybe that could be documented in some comment,

PR #8, opened 2021-08-24
Fix IPv4 not working properly on dual mode cellular

LaakkonenJussi:
Might not be the optimal solution but this resolves the issue. By adding an additional check to not to change the ipconfig method unnecessarily and not allowing to update modem ipconfiguration when disconnecting the issue seems to go away. Nevertheless, it was wrong to change ipconfig method from what ofono informed it to be. Utilizing the disconnecting check ensures that settings file is not updated with invalid information.

In addition to this a delayed connect for the cellular is implemented. This will ensure that all IP families do have proper configuration set. The checks apply only to FIXED and MANUAL mode configurations which do need to have local address set by ofono and informed to connmand. Otherwise in some bootup cases there will be two IPv4 addresses for the interface as the old saved IP address etc. are read from settings file as the information about IPv4 network comes late.

It seemed that rtnl.c does get the information beforehand but as ConnMan is built this way it just notifies the system set addresses forward. Messing up with that approach may break more than I'd aim to fix with it. Therefore, ofono should be aware when it can set the network connected. The approach waits for approx. 2 seconds until giving up - i.e., the address information should be arriving within that timeframe and in testing this is done after the first delay, so in error cases the delay in getting mobile data up is 200ms at most, 100ms being average.

Also took some DHCP patches from upstream.

### DIFF
--- a/connman/plugins/sailfish_ofono.c
+++ b/connman/plugins/sailfish_ofono.c
@@ -506,15 +506,21 @@ static GString *modem_append_strv(GString *str, char *const *strv)
 static GString *modem_configure_ipv4(struct connman_network *network,
 	const struct ofono_connctx_settings *config, GString *nameservers)
 {
-	if (config->method == OFONO_CONNCTX_METHOD_STATIC && config->address) {
-		struct connman_ipaddress *ipaddr =
+	DBG("config %p address %p dns %p", config, config->address,
+								config->dns);
+
+	if (config->method == OFONO_CONNCTX_METHOD_STATIC) {
+		if (config->address) {
+			struct connman_ipaddress *ipaddr =
 					connman_ipaddress_alloc(AF_INET);
-		connman_ipaddress_set_ipv4(ipaddr, config->address,
+			connman_ipaddress_set_ipv4(ipaddr, config->address,
 					config->netmask, config->gateway);
+			connman_network_set_ipaddress(network, ipaddr);
+			connman_ipaddress_free(ipaddr);
+		}
+
 		connman_network_set_ipv4_method(network,
 					CONNMAN_IPCONFIG_METHOD_FIXED);
-		connman_network_set_ipaddress(network, ipaddr);
-		connman_ipaddress_free(ipaddr);
 	} else {
 		connman_network_set_ipv4_method(network,
 					CONNMAN_IPCONFIG_METHOD_DHCP);
@@ -556,19 +562,24 @@ static int modem_configure(struct modem_data *md)
 		DBG("%s %d", ofono_modem_path(md->modem), index);
 
 		if (md->connctx->settings) {
+			DBG("IPv4 method %d", md->connctx->settings->method);
 			connman_service_create_ip4config(service, index);
 			ns = modem_configure_ipv4(md->network,
 					md->connctx->settings, ns);
 		} else {
+			DBG("set network %p IPv4 DHCP", md->network);
 			connman_network_set_ipv4_method(md->network,
 					CONNMAN_IPCONFIG_METHOD_DHCP);
 		}
 
 		if (md->connctx->ipv6_settings) {
+			DBG("IPv6 method %d",
+					md->connctx->ipv6_settings->method);
 			connman_service_create_ip6config(service, index);
 			ns = modem_configure_ipv6(md->network,
 					md->connctx->ipv6_settings, ns);
 		} else {
+			DBG("set network %p IPv6 AUTO", md->network);
 			connman_network_set_ipv6_method(md->network,
 					CONNMAN_IPCONFIG_METHOD_AUTO);
 		}
@@ -585,6 +596,9 @@ static int modem_configure(struct modem_data *md)
 static void modem_connected(struct modem_data *md)
 {
 	const int index = modem_configure(md);
+
+	DBG("index %d ipv4 %p ipv6 %p", index, md->connctx->settings,
+				md->connctx->ipv6_settings);
 
 	if (index >= 0) {
 		connman_network_set_index(md->network, index);
@@ -710,6 +724,17 @@ static void connctx_active_changed(OfonoConnCtx *connctx, void *arg)
 
 static void connctx_settings_changed(OfonoConnCtx *connctx, void *arg)
 {
+	struct modem_data *md = arg;
+	bool disconnecting = connman_network_get_disconnecting(md->network);
+
+	DBG("index %d ipv4 %p ipv6 %p", connman_network_get_index(md->network),
+			md->connctx->settings, md->connctx->ipv6_settings);
+
+	if (disconnecting) {
+		DBG("network %p disconnecting, skip modem conf", md->network);
+		return;
+	}
+
 	modem_configure(arg);
 }
 
--- a/connman/plugins/sailfish_ofono.c
+++ b/connman/plugins/sailfish_ofono.c
@@ -791,6 +791,27 @@ static void connctx_valid_changed(OfonoConnCtx *connctx, void *arg)
 	modem_update_network(arg);
 }
 
+static enum connman_ipconfig_type get_ofono_ipconfig_type(
+						struct modem_data *data)
+{
+	if (!data)
+		return CONNMAN_IPCONFIG_TYPE_UNKNOWN;
+
+	switch (data->connctx->protocol) {
+	case OFONO_CONNCTX_PROTOCOL_UNKNOWN:
+	case OFONO_CONNCTX_PROTOCOL_NONE:
+		return CONNMAN_IPCONFIG_TYPE_UNKNOWN;
+	case OFONO_CONNCTX_PROTOCOL_IP:
+		return CONNMAN_IPCONFIG_TYPE_IPV4;
+	case OFONO_CONNCTX_PROTOCOL_IPV6:
+		return CONNMAN_IPCONFIG_TYPE_IPV6;
+	case OFONO_CONNCTX_PROTOCOL_DUAL:
+		return CONNMAN_IPCONFIG_TYPE_ALL;
+	}
+
+	return CONNMAN_IPCONFIG_TYPE_UNKNOWN;
+}
+
 static void connctx_update_active(struct modem_data *md)
 {
 	GASSERT(md->connctx);
@@ -816,6 +837,8 @@ static void connctx_update_active(struct modem_data *md)
 				modem_clean_delayed_set_connected(md);
 				connman_network_set_connected(md->network,
 								FALSE);
+				connman_network_clear_ipaddress(md->network,
+						get_ofono_ipconfig_type(md));
 			}
 		}
 	}
@@ -834,7 +857,8 @@ static void connctx_settings_changed(OfonoConnCtx *connctx, void *arg)
 	struct modem_data *md = arg;
 	bool disconnecting = connman_network_get_disconnecting(md->network);
 
-	DBG("index %d ipv4 %p ipv6 %p", connman_network_get_index(md->network),
+	DBG("index %d ipv4 %p ipv6 %p",md->network ?
+				connman_network_get_index(md->network) : -1,
 			md->connctx->settings, md->connctx->ipv6_settings);
 
 	if (disconnecting) {
--- a/connman/plugins/sailfish_ofono.c
+++ b/connman/plugins/sailfish_ofono.c
@@ -98,6 +98,10 @@ enum connctx_handler_id {
 /* Should be less than CONNECT_TIMEOUT defined in service.c (currently 120) */
 #define ACTIVATE_TIMEOUT_SEC (60)
 #define ONLINE_CHECK_SEC  (2)
+/* Timeout for delayed set connected call, in milliseconds. */
+#define DELAYED_CONNECT_TIMEOUT_MS (100)
+/* Wait for approx 2s for address information from ofono. */
+#define DELAYED_CONNECT_LIMIT (2000 / DELAYED_CONNECT_TIMEOUT_MS)
 
 struct modem_data {
 	OfonoModem *modem;
@@ -121,6 +125,8 @@ struct modem_data {
 	guint strength;
 	char *name;
 	char *imsi;
+	guint delayed_set_connected_id;
+	guint delayed_set_connected_attempts;
 };
 
 struct plugin_data {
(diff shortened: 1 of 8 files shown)


LaakkonenJussi:
@inzanity review is appreciated.

connman/src/network.c
```diff
@@ -1538,6 +1549,7 @@ int __connman_network_disconnect(struct connman_network *network)
 		return -EUNATCH;
 
 	network->connecting = false;
+	network->disconnecting = true;
```
LaakkonenJussi, line 1552: Before anyone comments I totally agree that network.c would start to benefit a network state machine...

inzanity (commented):
Can't spot anything too problematic with it. Adding the disconnecting flag doesn't feel that nice, but if it works...

I guess the alternative would be to keep the flag inside the plugin, but that could hit own feet later on too.

LaakkonenJussi:
> Can't spot anything too problematic with it. Adding the disconnecting flag doesn't feel that nice, but if it works...
> 
> I guess the alternative would be to keep the flag inside the plugin, but that could hit own feet later on too.

I need to still test if that disconnecting flag is required but it just seems a bit wrong to configure anything when already disconnecting. And change in each, IPv4 and IPv6 do come as separate calls as well as the final one without any structures:
```
13:12:12.367 plugins/sailfish_ofono.c: ofono_network_disconnect() /ril_0 network 0x24d43ed0
13:12:12.367 Deactivating /ril_0/context1
13:12:12.407 plugins/sailfish_ofono.c: connctx_settings_changed() index 35 ipv4 0x24d40088 ipv6 0x24d400f8
13:12:12.408 plugins/sailfish_ofono.c: connctx_settings_changed() network 0x24d43ed0 disconnecting, skip modem conf
13:12:12.543 plugins/sailfish_ofono.c: connctx_settings_changed() index 35 ipv4 (nil) ipv6 0x24d400f8
13:12:12.543 plugins/sailfish_ofono.c: connctx_settings_changed() network 0x24d43ed0 disconnecting, skip modem conf
13:12:12.544 IPv6.Settings.PrefixLength: 0
13:12:12.545 Interface: <none>
13:12:12.545 plugins/sailfish_ofono.c: connctx_settings_changed() index 35 ipv4 (nil) ipv6 (nil)
13:12:12.546 plugins/sailfish_ofono.c: connctx_settings_changed() network 0x24d43ed0 disconnecting, skip modem conf
```

I was thinking of adding a state machine to `network.c` which would clarify things a lot but it is quite a laborous task. Tried to utilize that  `struct OfonoConnCtx active` but it seemed not to function as expected in this case as it is changed apparently later on. The state machine would make a lot of sense, though, it just needs to be kept in sync with `service.c`. Maybe that deserves an own task?

LaakkonenJussi:
Without the disconnecting check the settings file of the cellular apparently can be left with inconsistent information as DHCP is being set as the method even though FIXED is the correct one and used prior to disconnection of the cellular. Thus, I think that is actually needed.

connman/plugins/sailfish_ofono.c
```diff
@@ -98,6 +98,10 @@ enum connctx_handler_id {
 /* Should be less than CONNECT_TIMEOUT defined in service.c (currently 120) */
 #define ACTIVATE_TIMEOUT_SEC (60)
 #define ONLINE_CHECK_SEC  (2)
+/* In milliseconds */
+#define DELAYED_CONNECT_TIMEOUT (100)
```
monich, line 102: I suggest DELAYED_CONNECT_TIMEOUT_MS to make it clear which units are used

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
...
+		if (md->delayed_connect_id)
+			g_source_remove(md->delayed_connect_id);
+
+		DBG("%p init delayed connect, modem network not ready", md);
+
+		md->delayed_connect_attempts = 0;
+		md->delayed_connect_id = g_timeout_add(DELAYED_CONNECT_TIMEOUT,
+							delayed_connect, md);
+	} else {
+		connman_network_set_connected(md->network, TRUE);
```
monich, line 684: Cancel the pending delayed config when it's no longer needed

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
...
+	md->delayed_connect_attempts = 0;
+
+	return FALSE;
+}
+
+static void modem_set_connected(struct modem_data *md)
+{
+	if (!is_modem_network_configured(md)) {
+		if (md->delayed_connect_id)
+			g_source_remove(md->delayed_connect_id);
```
monich, line 676: Would be nice to have `{}` around the if blocks consistently

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
...
+
+		break;
+	}
+
+	DBG("%p is configured", md);
+
+	return TRUE;
+}
+
+static gboolean delayed_connect(gpointer data) {
```
monich, line 647: How about naming it `modem_delayed_connect()` or something, to reduce the number of prefixes used in this file. And the opening bracket should be on the next line.

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
 	return index;
 }
 
+static gboolean is_modem_network_configured(struct modem_data *md)
```
monich, line 602: This one I would call `modem_is_network_configured()`

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
...
+	case OFONO_CONNCTX_PROTOCOL_UNKNOWN:
+	case OFONO_CONNCTX_PROTOCOL_NONE:
+		return FALSE;
+	case OFONO_CONNCTX_PROTOCOL_IP:
+		if (!connman_network_is_configured(md->network,
+						CONNMAN_IPCONFIG_TYPE_IPV4)) {
+			connman_warn("ofono: %p IPv4 no address set", md);
+			return FALSE;
+		}
+
```
monich, line 616: Quite a few empty lines in this function could be deleted to make it more compact. This file is not unstreamable and therefore can afford to be more readable than the rest of the connman code)) Feel free to ignore though.

connman/plugins/sailfish_ofono.c
```diff
@@ -121,6 +125,8 @@ struct modem_data {
 	guint strength;
 	char *name;
 	char *imsi;
+	guint delayed_connect_id;
```
monich, line 128: I would consider naming it `delayed_set_connected`, `delayed_connect_notify` or something. I was confused by this naming.

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
...
+	g_source_remove(md->delayed_connect_id);
+	md->delayed_connect_id = 0;
+	md->delayed_connect_attempts = 0;
+
+	return FALSE;
+}
+
+static void modem_set_connected(struct modem_data *md)
+{
+	if (!is_modem_network_configured(md)) {
```
monich, line 674: I think swapping the if blocks would improve the readability - if (network is configured) { do this } else { do something else }

connman/plugins/sailfish_ofono.c
```diff
@@ -593,6 +599,92 @@ static int modem_configure(struct modem_data *md)
...
+}
+
+static gboolean delayed_connect(gpointer data) {
+	struct modem_data *md = data;
+
+	/* Keep in loop until configured and attempt limit is not reached. */
+	if (!is_modem_network_configured(md) &&
+			md->delayed_connect_attempts < DELAYED_CONNECT_LIMIT) {
+		md->delayed_connect_attempts++;
+		return TRUE;
```
inzanity, line 654: I'd prefer the use of G_SOURCE_CONTINUE (and _REMOVE) for more self-documenting code.

connman/plugins/sailfish_ofono.c
```diff
@@ -593,6 +599,92 @@ static int modem_configure(struct modem_data *md)
...
+
+	/* Log that we're giving up, network setup wasn't completed in time. */
+	if (md->delayed_connect_attempts == DELAYED_CONNECT_LIMIT)
+		connman_error("cellular setup was not completed in time");
+
+	DBG("modem %p network %p configured, set connected", md, md->network);
+
+	connman_network_set_connected(md->network, TRUE);
+
+	g_source_remove(md->delayed_connect_id);
```
inzanity, line 665: The return FALSE (or G_SOURCE_REMOVE) makes explicit removal unnecessary.

connman/plugins/sailfish_ofono.c
```diff
@@ -695,6 +801,12 @@ static void connctx_update_active(struct modem_data *md)
 			if (md->network) {
 				connman_network_set_connected(md->network,
 								FALSE);
+				if (md->delayed_connect_id) {
```
monich, line 804: I would do the cleanup before calling `connman_network_set_connected()`, and generally try be in a consistent state when calling any connman core function. Just in case if connman feels like calling me back before returning from the function.

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
...
+}
+
+static gboolean delayed_connect(gpointer data) {
+	struct modem_data *md = data;
+
+	/* Keep in loop until configured and attempt limit is not reached. */
+	if (!is_modem_network_configured(md) &&
+			md->delayed_connect_attempts < DELAYED_CONNECT_LIMIT) {
+		md->delayed_connect_attempts++;
+		return TRUE;
```
monich, line 654: Please `G_SOURCE_CONTINUE` for readability.

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
...
+
+	DBG("modem %p network %p configured, set connected", md, md->network);
+
+	connman_network_set_connected(md->network, TRUE);
+
+	g_source_remove(md->delayed_connect_id);
+	md->delayed_connect_id = 0;
+	md->delayed_connect_attempts = 0;
+
+	return FALSE;
```
monich, line 669: `G_SOURCE_REMOVE`

connman/plugins/sailfish_ofono.c
```diff
@@ -582,13 +599,102 @@ static int modem_configure(struct modem_data *md)
...
+
+	/* Log that we're giving up, network setup wasn't completed in time. */
+	if (md->delayed_connect_attempts == DELAYED_CONNECT_LIMIT)
+		connman_error("cellular setup was not completed in time");
+
+	DBG("modem %p network %p configured, set connected", md, md->network);
+
+	connman_network_set_connected(md->network, TRUE);
+
+	g_source_remove(md->delayed_connect_id);
```
monich, line 665: Again, it's safer to be in a consistent state when calling a connman function, i.e. complete the cleanup before `connman_network_set_connected()` call.

connman/plugins/sailfish_ofono.c
monich: Hehe, Santtu was there first))

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Well, it is the `id` of the event source and for the (modem_)`delayed_connect()` so.. for me "notify" would in ConnMan context indicate something to do with the notifier functionality.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Changed,

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Added a function for this as it is to be done in similar fashion when disconnecting as well. When this is called we should anyways clean those values to either connect or start the delayed connect.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Moved to other function where braces are applied.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Yeap, makes it more readable. Changed.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: True, follows the other code nicely. Changed,

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Added some and I guess I got your drift here.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Yep, done.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Yep, done.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Yep, that was unnecessary indeed. Apparently I was too focused on the issue at hand than at the code.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Done, added a new function for that to be a lot cleaner.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Yup, done.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Yup, done.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: This was unnecessary anyways. When returning `G_SOURCE_REMOVE` it is naturally handled.

monich: Not just unnecessary, it was actually wrong (and I didn't catch that)! It would result in deleting the same source twice.

LaakkonenJussi: The Finnish phrase "Rapatessa roiskuu" works perfectly here ;)j

connman/plugins/sailfish_ofono.c
```diff
@@ -662,26 +665,34 @@ static gboolean delayed_connect(gpointer data) {
...
 
-	g_source_remove(md->delayed_connect_id);
 	md->delayed_connect_id = 0;
 	md->delayed_connect_attempts = 0;
 
-	return FALSE;
+	return G_SOURCE_REMOVE;
+}
+
+static void clean_modem_delayed_connect(struct modem_data *md)
```
inzanity: I think clean_ is somewhat ambiguous here (at least I interpret clean primarily as an adjective, maybe I'm a lazy cleaner...), I'd prefer clear_ or cleanup_, but then again, naming of internal functions is not that important.

LaakkonenJussi: Yeah maybe `modem_clear_delayed_connect()` sounds better anyways? Changed to use that.

connman/plugins/sailfish_ofono.c
```diff
@@ -671,7 +672,7 @@ static gboolean modem_delayed_connect(gpointer data) {
 	return G_SOURCE_REMOVE;
 }
 
-static void clean_modem_delayed_connect(struct modem_data *md)
+static void modem_clear_delayed_connect(struct modem_data *md)
```
inzanity, line 675: If not too much trouble, could you move these renames to the commit that introduces this function, would be less noise that way.

inzanity (approved):
LGTM, a few changes could be moved around, but not necessary.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Done.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: Renamed further to be in line with the rest.

connman/plugins/sailfish_ofono.c
LaakkonenJussi: I guess `delayed_set_connected_id` is the compromise here.

PR #10, opened 2021-09-13
Create StorageRoot with 755 if it does not exist and document options.

LaakkonenJussi:
Change connmand behavior to utilize different permissions for the StorageRoot creation if it does not exist to allow other users to view the contents as well. The subdirectories `connman` and `connman-vpn` are created using stricter permissions (root only). Default to 755 permissions with StorageRoot.

Document the StorageRoot related permissions into conf man page.


spiiroin (approved):
LGTM

PR #12, opened 2021-09-17
Prevent symlinks and references in paths.

LaakkonenJussi:
Prevent using a path that contains symlinks or references by checking
that the path expanded with realpath() matches to the original. This way
we can be sure that the location defined in settings is not actually
pointing to any other directory, e.g., via a symlink to another users'
files.

When the paths match or the path does not exist yet storage.c functions
as normal. In case the path is unusable we continue working but cannot
save anything to the location. This applies to any level in the path, so
if a user attempts to symlink a single service dir or a file into the
users' save path even using root user the use of such content is
prevented.


connman/src/storage.c
```diff
@@ -794,11 +796,58 @@ static void keyfile_cleanup(void)
...
+	}
+
+	/*
+	 * If the paths differ the path most likely contains symlink or is not
+	 * an absolute path.
+	 */
+	if (g_strcmp0(pathname, expanded_path))
+		err = -ENOTSUP;
+
+	g_free(expanded_path);
```
inzanity, line 832: expanded_path doesn't come from a GLib function, so normal free() should be used instead.

LaakkonenJussi: Yeah, true. Changed and force pushed.

connman/src/storage.c
```diff
@@ -794,11 +796,58 @@ static void keyfile_cleanup(void)
...
+		case EINVAL: /* pathname was NULL */
+		case EIO:
+		case ELOOP:
+		case ENAMETOOLONG:
+		case ENOMEM:
+		case ENOTDIR:
+			DBG("realpath() error %d/%s", errno, strerror(errno));
+			return -errno;
+		/* When saving the file does not exist */
+		case ENOENT:
```
inzanity, line 818: Note: ENOENT will be returned for symlink-to-missing. Not sure if that is an issue, but I think you could at least store data in another user's data dir.

connman/src/storage.c
```diff
@@ -794,11 +796,58 @@ static void keyfile_cleanup(void)
...
+}
+
 static GKeyFile *storage_load(const char *pathname)
 {
 	struct keyfile_record *record = NULL;
 	GError *error = NULL;
 	GKeyFile *keyfile = NULL;
+	int err;
+
+	err = validate_path(pathname);
```
inzanity, line 845: Does it make sense to do this before the cache check?

inzanity:
I think this improves things, although it still leaves a short race between validation and opening. Not quite sure what would be the atomic version of this, probably something like:
```
int dfd = open("/", O_DIRECTORY | O_NOFOLLOW | O_RDONLY);
int fd;
char **parts = g_strsplit(pathname, "/", -1);
size_t i;
for (i = 1; parts[i + 1]; i++) {
    int sdfd = openat(dfd, parts[i], O_DIRECTORY | O_NOFOLLOW | O_RDONLY);
    if (sdfd < 0)
        goto out;
    close(dfd);
    dfd = sdfd;
}
fd = openat(dfd, parts[i], O_RDONLY);
/* read all data and parse it with g_key_file_load_from_data (or _bytes) */
close(fd);
out:
close(dfd);
g_strfreev(parts);

return something;
```

connman/src/storage.c
LaakkonenJussi: [man page](https://www.man7.org/linux/man-pages/man3/realpath.3.html) only states that `ENOENT The named file does not exist.` Are you referring to the:
```
GNU extensions
       If the call fails with either EACCES or ENOENT and resolved_path
       is not NULL, then the prefix of path that is not readable or does
       not exist is returned in resolved_path.
```
In which case `errno` should be checked also when the `expanded_path` is set.

connman/src/storage.c
LaakkonenJussi: Maybe not now that you brought that up. Late Friday changes are at first like this. It is indeed faster and one allocation less in cases where the service-files etc. are not user specific as this issue exist only when multiuser support is in use.

LaakkonenJussi:
> I think this improves things, although it still leaves a short race between validation and opening. Not quite sure what would be the atomic version of this, probably something like:
> 
> ```
> int dfd = open("/", O_DIRECTORY | O_NOFOLLOW | O_RDONLY);
> int fd;
> char **parts = g_strsplit(pathname, "/", -1);
> size_t i;
> for (i = 1; parts[i + 1]; i++) {
>     int sdfd = openat(dfd, parts[i], O_DIRECTORY | O_NOFOLLOW | O_RDONLY);
>     if (sdfd < 0)
>         goto out;
>     close(dfd);
>     dfd = sdfd;
> }
> fd = openat(dfd, parts[i], O_RDONLY);
> /* read all data and parse it with g_key_file_load_from_data (or _bytes) */
> close(fd);
> out:
> close(dfd);
> g_strfreev(parts);
> 
> return something;
> ```

I thought of that first but sought out a simpler already implemented solution...

connman/src/storage.c
inzanity: The man page didn't give enough information, so I created a trivial test case to see how it works :)

The GNU extension can indeed be used to check for the symlink-to-missing, but you'll need to use the buffer argument to get the info.

So with a symlink-to-missing, you'll get: rv: NULL, buffer will be filled with the name of the linked file, errno ENOENT

LaakkonenJussi: Oh joy, world shall burn when you cannot trust man pages ;)

LaakkonenJussi: Actually it seems that `g_file_set_contents()` does fail if there is a symlink to missing location for a service about to be saved in `storage_save()`. But anyways, made a reformatting to be more clear in call cases (hopefully).

connman/src/storage.c
LaakkonenJussi: Changed the approach so `g_free()` became valid after all :).

connman/src/storage.c
```diff
@@ -794,11 +796,71 @@ static void keyfile_cleanup(void)
 	keyfile_hash = NULL;
 }
 
+static int validate_path(const char *pathname)
+{
+	char *resolved_path;
+	char *ret;
+	int err = 0;
+
+	resolved_path = g_try_malloc0(PATH_MAX);
```
inzanity, line 805: Why not just char resolved_path[PATH_MAX]?

LaakkonenJussi: Yeah, less mem fragmentation. Removed also use of that `char *ret` as it is not actually needed anymore.

connman/src/storage.c
```diff
@@ -794,11 +796,60 @@ static void keyfile_cleanup(void)
...
+
+	errno = 0;
+
+	/*
+	 * realpath() expands all symbolic links and references. Returns
+	 * ENOENT if pathname is NULL.
+	 */
+	if (!realpath(pathname, resolved_path)) {
+		/* File does not exist when saving, return 0.*/
+		if (errno == ENOENT)
```
inzanity, line 811: This isn't quite enough, you should still check if resolved_path matches pathname in this case, otherwise this can still be used to create new configs in another user's dirs with symlinks.

connman/src/storage.c
```diff
@@ -794,11 +796,60 @@ static void keyfile_cleanup(void)
...
+			return 0;
+
+		DBG("realpath() error %d/%s", errno, strerror(errno));
+		return -errno;
+	}
+
+	switch (errno) {
```
spiiroin, line 825: What is this about? AFAICT the errors that realpath() intentionally returns have already been handled above and this would cases where realpath() has returned success but errno has still changed -> just noise internal to realpath() that should not be of interest to anybody, or am I missing something?

connman/src/storage.c
LaakkonenJussi: Yes when this now uses `resolved_path`, added a commit 4b2bc8804c946fdaa4feb0b756859db39cbe726b to address this and cleaned up some bits.

connman/src/storage.c
LaakkonenJussi: It is the GNU extension of `realpath()`:￼ ￼￼https://github.com/sailfishos/connman/pull/12#discussion_r711955362 as it apparently is not documented well enough it can set error with a non-NULL return in case there is a broken symlink. Or are we missing something?

spiiroin: "If the call fails with either EACCES or ENOENT and resolved_path is not NULL..."

To me that reads as:

    if(!realpath(foo,bar) && (errno==EACCESS || errno==ENOENT)) gnu_stuff_happened;

and nothing points towards errno having any meaningful value when realpath() succeeds i.e. non-null value is returned

spiiroin: iow I still think you should get rid of this "check errno after success" block (and possibly also that earlier errno=0 while at it).

connman/src/storage.c
```diff
@@ -794,11 +796,53 @@ static void keyfile_cleanup(void)
...
+		switch (errno) {
+		/*
+		 * File does not exist, which is the case when saving. Verify
+		 * the path to avoid using a broken symlink or unaccessible
+		 * file.
+		 */
+		case ENOENT:
+			DBG("path %s is not readable/has missing symlink (%s)",
+						pathname, resolved_path);
+			break;
```
spiiroin, line 817: I wonder should we care that 'resolved_path' content is undefined here / after break unless we are dealing with realpath() with gnu extensions? Is it even possible to know which way it is?

connman/src/storage.c
```diff
@@ -794,11 +796,53 @@ static void keyfile_cleanup(void)
...
+		case EACCES:
+		case EIO:
+		case ELOOP:
+		case ENAMETOOLONG:
+		case ENOMEM:
+		case ENOTDIR:
+		default:
+			DBG("realpath() error %d/%s", errno, strerror(errno));
```
spiiroin, line 825: Logging also 'pathname' might be useful from debugging point of view.

spiiroin (approved):

connman/src/storage.c
inzanity: That is a reasonably good question; musl's realpath() doesn't seem to implement the GNU extension.

spiiroin: What might be sufficiently close to gnu ENOENT behavior we are after could be something like:
if realpath(parent(pathname)) succeeds, append basename(pathname) back in and do access()/stat()/whatnot checks on the result (in case it happens to be something weird like a broken link).

Though given that afaik we are mostly interested in not having symlinks on the way, that openat() loop starts to look like something simple and with well defined behavior to me...

LaakkonenJussi: Files are saved surprisingly often so with every file save that allocation caused by splitting of the path seems to be a bit too much. Anyways, `g_file_set_contents()` does not allow saving a file to a path with a broken symlink...

connman/src/storage.c
LaakkonenJussi: Yep got a bit confused here. Those have been changed now.

connman/src/storage.c
LaakkonenJussi: Calling functions do printout a warning in that case when there is an error.

connman/src/storage.c
spiiroin: Well, I'd assume internally realpath() is basically something like that and then some more, so in all likelihood custom solution that bails out the moment there is a symlink imight actually be the lighter option.

But for our purposes relying on gnu extended realpath() works, so I'll leave it up to you whether stricter standards compliance is needed or not.

connman/src/storage.c
spiiroin: ok, so it seems.

spiiroin (approved):

connman/src/storage.c
inzanity: Yeah, this is by far the simplest implementation, but I suppose the performance difference is very minor (esp. if you implement the splitting with strtok/similar instead of g_strsplit).

If one wants to go with the non-racy version, that would require re-implementing quite a few helpers (even mkstemp() doesn't have a *at() version).

I'm cool with current implementation, even with the slight race it leaves behind.

inzanity (approved):

PR #13, opened 2021-09-29
technology: Create settings file in user change post command

LaakkonenJussi:
[technology] Create settings file in user change post command. JB#55593

Add connman_technology_save_offlinemode() to the user change post
command (__connman_technology_enable_from_config()) to create an empty
settings file if it does not exist similar to
__connman_technology_init() behavior. This ensures that after every user
change the user has at least empty settings file with the current
offline mode set. The offline mode is in this case disabled (set to
"false") as new users should begin in online mode.


connman/src/technology.c
```diff
@@ -1946,6 +1982,17 @@ bool __connman_technology_enable_from_config()
 
 	keyfile = __connman_storage_load_global();
 	if (!keyfile) {
+		/*
+		 * When the settings file does not exist create it similarly to
+		 * technology is initialization. Since this concerns new
+		 * users set the global offlinemode to "false" as new users
+		 * should not have offline mode set.
+		 */
+		if (global_offlinemode)
```
inzanity, line 1991: `_set_offlinemode()` already does this check, so you could just call it directly here. Also, I'd prefer using `false` instead of the variable. (one less thing to remember)

connman/src/technology.c
```diff
@@ -1037,6 +1044,23 @@ static DBusMessage *set_powered(struct connman_technology *technology,
...
 
+	return err;
+}
+
+
+static DBusMessage *set_powered(struct connman_technology *technology,
+				DBusMessage *msg, bool powered)
+{
+	DBusMessage *reply = NULL;
+	int err = 0;
```
inzanity, line 1055: This is just copy-paste, but the initialization here violates coding style M7.

connman/src/technology.c
```diff
@@ -49,6 +49,7 @@ static bool global_offlinemode;
 static unsigned int global_offlinemode_override; /* Technology bitmask */
 struct connman_access_tech_policy *tech_access_policy;
 static unsigned int enable_delayed_ids[MAX_CONNMAN_SERVICE_TYPES] = { 0 };
+static bool save_changes = true;
```
inzanity, line 52: I'm not objecting to this, but just thinking would it be better to have some kind of "changing user" flag instead.

inzanity (approved):
Looks reasonable to me.

LaakkonenJussi:
> Looks reasonable to me.

Actually this was a mere PoC that there is something wrong at the UI level :D. It would seem to be the lib that is not sending the powered on call to connmand as there are no technology enabled because that powered on hasn't been sent. The problem lies elsewhere and I think this is unnecessary. Sorry, old git.sailfishos.org I think set the whole MR as WIP if there was at least one commit with that prefix..

aroditelev (approved):

PR #15, opened 2021-10-12
[packaging] Only BuildRequire systemd via pkgconfig. JB#55010

Thaodan:
We already include systemd as BuildRequires lets drop it so the right systemd
variant gets installed into the builder.


xfade (approved):
I think this will work otherwise it would have failed to build in the test project with systemd-mini. (if it really needed those extra dependencies on build time)

PR #18, opened 2021-10-14
Improve OpenVPN config data processing and reset handling

LaakkonenJussi:
Add a config option type that is providing better granularity in how the options should be processed when adding to the task, which eventually sets them to the binary. Also support the script context in both OpenVPN script and plugin, as of now simply stop the task and make the VPN disconnected if reset. The handling of VPN reset is to be addressed in later work.

### DIFF
--- a/connman/vpn/vpn-provider.c
+++ b/connman/vpn/vpn-provider.c
@@ -1603,6 +1603,7 @@ static void connect_cb(struct vpn_provider *provider, void *user_data,
 						VPN_PROVIDER_STATE_DISCONNECT);
 			}
 			break;
+		case ECONNRESET:
 		default:
 			vpn_provider_indicate_error(provider,
 					VPN_PROVIDER_ERROR_CONNECT_FAILED);


connman/vpn/plugins/openvpn.c
```diff
@@ -357,38 +366,52 @@ static int ov_save(struct vpn_provider *provider, GKeyFile *keyfile)
 static int task_append_config_data(struct vpn_provider *provider,
 					struct connman_task *task)
 {
-	const char *option;
+	const char *option = NULL;
```
inzanity, line 369: This should be moved inside the loop; now string values can leak to bool side.

LaakkonenJussi: Yep, forgot that completely when changing the arg adding part. Thanks.

inzanity: My thinking was the move the whole variable there, not used outside the loop block; this works too though.

LaakkonenJussi: Well, now you mentioned it, that looks even better. Updated the review fix commit.

connman/vpn/plugins/openvpn.c
```diff
@@ -357,38 +366,52 @@ static int ov_save(struct vpn_provider *provider, GKeyFile *keyfile)
...
+		switch (ov_options[i].opt_type) {
+		case OPT_NONE:
 			continue;
 
-		option = vpn_provider_get_string(provider,
-					ov_options[i].cm_opt);
-		if (!option)
-			continue;
+		case OPT_STRING:
+			if (!ov_opt)
```
inzanity, line 382: This seems a bit redundant now with the addition of the OPT_NONE. No real harm with keeping it though.

LaakkonenJussi: I was thinking whether to use it or not. Especially since `connman_task_add_argument()` does check that name exists and returns error if it does not -> `task_append_config_data()` would return in that case.. Maybe this could be left to avoid misuse in `ov_options[]`.

PR #19, opened 2021-11-09
Fix USB ethernet hot-plug

LaakkonenJussi:
This change fixes the re-insertion of a hot-plug USB ethernet. The index of the interface was not updated so after re-insertion the old index was used which resulted in DHCP failing to start. Also included one fix from upstream for ethernet.

### DIFF
--- a/connman/plugins/ethernet.c
+++ b/connman/plugins/ethernet.c
@@ -73,7 +73,7 @@ static int get_vlan_vid(const char *ifname)
 		return -errno;
 
 	vifr.cmd = GET_VLAN_VID_CMD;
-	stpncpy(vifr.device1, ifname, sizeof(vifr.device1));
+	stpncpy(vifr.device1, ifname, sizeof(vifr.device1) - 1);
 
 	if(ioctl(sk, SIOCSIFVLAN, &vifr) >= 0)
 		vid = vifr.u.VID;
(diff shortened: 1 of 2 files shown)


connman/src/service.c
```diff
@@ -9343,8 +9343,19 @@ bool __connman_service_create_from_network(struct connman_network *network)
 	if (__connman_network_get_weakness(network))
 		return true;
 
+	index = connman_network_get_index(network);
+
 	if (service->path) {
 		update_from_network(service, network);
+
```
inzanity, line 9350: Is there some reason we can't do these if-sets right after getting the index, then we wouldn't need to have the same code twice?

connman/plugins/ethernet.c
```diff
@@ -73,7 +73,7 @@ static int get_vlan_vid(const char *ifname)
 		return -errno;
 
 	vifr.cmd = GET_VLAN_VID_CMD;
-	stpncpy(vifr.device1, ifname, sizeof(vifr.device1));
+	stpncpy(vifr.device1, ifname, sizeof(vifr.device1) - 1);
```
inzanity, line 76: Ed. note: I think this is a wrong fix to a non-issue (kernel does null terminate these), but it's from upstream, so let's just roll with it. IMO correct fix would be to err out if the name does not fit in the buffer.

inzanity (approved):
LGTM, but left a note for potential simplification.

connman/plugins/ethernet.c
LaakkonenJussi: Yeah... well, we can still drop this and propose a proper fix to upstream (?) as it does have the same issue regarding hot-plugs afaik.

LaakkonenJussi: Or, well, lets keep it the same as upstream. We'll fix it later if this causes  any issues, not that it really should but.

PR #20, opened 2021-11-22
[ipconfig] Do not enable/disable ipv6 for all ifs. Fixes JB#56478

LaakkonenJussi:
This is a patch proposed to upstream. And this is indeed an issue, since the USB device is already removed prior processing this bits in ifconfig.c -> IPv6 gets disabled. This was an unseen and surprising side effect of how USB devices are reacted to.

### DIFF
--- a/connman/src/ipconfig.c
+++ b/connman/src/ipconfig.c
@@ -1736,6 +1736,9 @@ static int disable_ipv6(struct connman_ipconfig *ipconfig)
 
 	ifname = connman_inet_ifname(ipconfig->index);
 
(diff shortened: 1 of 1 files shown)


LaakkonenJussi:
Not yet approved to upstream: https://lore.kernel.org/connman/20211122130346.114768-1-christian.taedcke@lemonbeat.com/ but does work well. Maybe a quick test tomorrow would not hurt with our VPN IPv6 disabling.

inzanity (approved):
Looks straightforward enough.

PR #21, opened 2021-11-25
Support cellular network dual mode according to 3GPP TS 24.301

LaakkonenJussi:
This improves the earlier fix for IPv4 not being enabled on cellular: https://github.com/sailfishos/connman/pull/8 by allowing to connect a dual mode network with only either of the IPv4 or IPv6 configured as it is stated in 3GPP TS 24.301: 
> if the UE requests for PDN type IPv4v6, but the subscription is limited to IPv4 only or IPv6 only for the requested APN, the network shall override the PDN type requested by the UE to be limited to a single address PDN type (IPv4 or IPv6).

### DIFF
--- a/connman/plugins/sailfish_ofono.c
+++ b/connman/plugins/sailfish_ofono.c
@@ -647,6 +647,33 @@ static gboolean modem_is_network_configured(struct modem_data *md)
 	return TRUE;
 }
 
+static void modem_ensure_dual_mode_configuration(struct modem_data *md)
+{
+	bool ipv4_configured;
+	bool ipv6_configured;
+
+	ipv4_configured = connman_network_is_configured(md->network,
+						CONNMAN_IPCONFIG_TYPE_IPV4);
+	ipv6_configured = connman_network_is_configured(md->network,
+						CONNMAN_IPCONFIG_TYPE_IPV6);
+
+	/* When both are unconfigured connection must fail */
+	if (ipv4_configured == ipv6_configured)
+		return;
+
+	if (!ipv4_configured) {
+		DBG("%p set IPv4 OFF", md);
+		connman_network_set_ipv4_method(md->network,
+						CONNMAN_IPCONFIG_METHOD_OFF);
+	}
+
+	if (!ipv6_configured) {
+		DBG("%p set IPv6 OFF", md);
+		connman_network_set_ipv6_method(md->network,
+						CONNMAN_IPCONFIG_METHOD_OFF);
+	}
+}
+
 static gboolean modem_delayed_set_connected(gpointer data)
 {
 	struct modem_data *md = data;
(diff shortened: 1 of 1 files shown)


connman/src/network.c
```diff
@@ -78,6 +78,7 @@ struct connman_network {
 	bool connecting;
 	bool associating;
 	bool disconnecting;
+	bool require_all_ip_types;
```
LaakkonenJussi, line 81: This could be renamed to be something better?

monich:
Missing IPv4 in dual mode is not a failure.

3GPP TS 24.301

> - if the UE requests for PDN type IPv4v6, but the subscription is limited to IPv4 only or IPv6 only for the
requested APN, the network shall override the PDN type requested by the UE to be limited to a single address
PDN type (IPv4 or IPv6).

LaakkonenJussi:
> Missing IPv4 in dual mode is not a failure.
> 
> 3GPP TS 24.301
> 
> > * if the UE requests for PDN type IPv4v6, but the subscription is limited to IPv4 only or IPv6 only for the
> >   requested APN, the network shall override the PDN type requested by the UE to be limited to a single address
> >   PDN type (IPv4 or IPv6).

Oh, that is good to know. So this is expected to happen. Good to know!

connman/src/network.c
```diff
@@ -626,8 +629,17 @@ static void set_connected(struct connman_network *network)
 	DBG("service %p ipv4 %p ipv6 %p", service, ipconfig_ipv4,
 		ipconfig_ipv6);
 
-	__connman_network_enable_ipconfig(network, ipconfig_ipv4);
-	__connman_network_enable_ipconfig(network, ipconfig_ipv6);
+	ipv4_err = __connman_network_enable_ipconfig(network, ipconfig_ipv4);
+	ipv6_err = __connman_network_enable_ipconfig(network, ipconfig_ipv6);
+
+	if (!network->require_all_ip_types) {
+		if (ipv4_err & ipv6_err)
```
inzanity, line 636: Probably meant to use && here?

LaakkonenJussi: Yeah! Good oops.

LaakkonenJussi: Included this in the network.c commit 5e2cd5dbb359f5ce752ef51ac89ae1eb594962ef

connman/src/network.c
```diff
@@ -1473,6 +1486,15 @@ bool connman_network_is_configured(struct connman_network *network,
 	return false;
 }
 
+void connman_network_set_require_all_ip_types(struct connman_network *network,
```
LaakkonenJussi, line 1489: I thought for this a bit because it is either setting a status variable like this or pass some appropriate `bool`/`enum connman_ipconfig_type` to define that what are required -> which would need for this case a new `enum` value. Latter option would need a change to the public plugin APIs and a lot of places. Any thoughts @inzanity @monich ?

inzanity (approved):
Don't see anything major wrong with this, just thinking would it be possible to set the ipconfig method to OFF if there is no configuration; that way the changes could stay inside the plugin altogether.

connman/src/network.c
LaakkonenJussi: Yeah, that was almost the same other option I had in mind for this. To deny or to allow.

LaakkonenJussi:
> Don't see anything major wrong with this, just thinking would it be possible to set the ipconfig method to OFF if there is no configuration; that way the changes could stay inside the plugin altogether.

That actually seems to work just as well. I'll reformat this to drop all the other changes and keep it within Ofono plugin.

LaakkonenJussi:
@inzanity Simplified this a lot. Thanks for the note. I did overthink and over-engineer that bit I guess.

connman/src/network.c
LaakkonenJussi: Obsolete. Resolving.

connman/src/network.c
LaakkonenJussi: This is gone now. Resolving.

inzanity (approved):
LGTM.

PR #22, opened 2021-12-01
Add heurestic for weak WiFi networks.

LaakkonenJussi:


connman/plugins/sailfish_wifi.c
```diff
@@ -638,7 +643,23 @@ static void wifi_bss_free(struct wifi_bss *bss_data)
...
 static inline gboolean wifi_bss_weak(const struct wifi_bss *bss_data)
 {
-	return (bss_data->signal < WIFI_WEAK_RSSI);
+	int weak_rssi = WIFI_WEAK_RSSI;
+	int services;
+
+	services = connman_service_get_service_count(CONNMAN_SERVICE_TYPE_WIFI);
+	if (services) {
+		int multiplier = (int)floor(
+					services / WIFI_SERVICE_COUNT_INTERVAL);
```
monich, line 652: Isn't that equivalent to just `multiplier = services / WIFI_SERVICE_COUNT_INTERVAL`?

LaakkonenJussi: I may have been overthinking here, again, and as a part of older approach I tried. Simplified with 227e5a58c5a82057284f23d68820568f467a864f

connman/plugins/sailfish_wifi.c
```diff
@@ -42,6 +42,7 @@
 #include <errno.h>
 #include <sys/socket.h>
 #include <linux/if.h>
+#include <math.h>
```
inzanity, line 45: Drop this too now that floor() is no longer used.

LaakkonenJussi:
@monich As the max number of BSS's is 100 should we make this to reach the BGSCAN limit of -65 dBm before we reach that number, so, lets say, increase with 5 dBm every 20 networks? Would it be too harsh? I guess you've worked on this a lot more than I have so you may have better insight on this. 30 is kind of safe I think.

connman/plugins/sailfish_wifi.c
LaakkonenJussi: True. Fixed with 15bb6a3b2a12ffee61392387acb586a97a807f98

inzanity:
Should we be counting services or networks (or services with networks)? Stored services shouldn't affect the number of items shown. Or have I totally forgotten how connman loads services...

LaakkonenJussi:
> Should we be counting services or networks (or services with networks)? Stored services shouldn't affect the number of items shown. Or have I totally forgotten how connman loads services...

AFAIK I can understand the WiFi code all of those BSS's that are within acceptable criteria based on signal and weak sightings are added to service.c as new unsaved networks. Yeah, that point about stored services might be good to account for.. that requires a bit of thinking to get it right. So should the BSS count be counted only by the WiFi plugin instead?

LaakkonenJussi:
> Should we be counting services or networks (or services with networks)? Stored services shouldn't affect the number of items shown. Or have I totally forgotten how connman loads services...

Might this be correct way of counting, at least the scope is now only on available ones: fbe46483886a45d23d15c8394345a2b8273f614b

LaakkonenJussi:
Ok, had to use hash tables to make sure none of the types are counted more than once. This is now a working version, @inzanity  @monich any comments? The increases now happen every 20 networks to the weak BSS signal strenght.

LaakkonenJussi:
Also @okodron review is appreciated.

okodron:
> Also @okodron review is appreciated.

CC @ikamaletdinov as well.

connman/src/service.c
```diff
@@ -510,10 +512,49 @@ static inline char *service_path(const char *ident)
...
+	DBG("%s service count: %d", __connman_service_type2string(type),
+					table ? g_hash_table_size(table) : 0);
+}
+
 static void service_remove(struct connman_service *service)
 {
 	service_list = g_list_remove(service_list, service);
 	g_hash_table_remove(service_hash, service->identifier);
+	g_hash_table_remove(service_type_hash[service->type],
+							service->identifier);
```
spiiroin, line 557: Noting that this does not guard agains service_type_hash[service->type] == NULL - why not use count_available_service_type(service, false)?

connman/src/service.c
```diff
@@ -9293,14 +9346,19 @@ static void update_from_network(struct connman_service *service,
...
 		service_schedule_updated(service);
-        }
+	}
 
-	if (was_available != service_available.value(service))
+	if (was_available != service_available.value(service)) {
 		service_boolean_changed(service, &service_available);
 
+		if (network_update)
+			count_available_service_type(service, true);
```
spiiroin, line 9359: Looks like service_available.value(service) is is_available(service) and count_available_service_type(service, true) is nop if is_available(service) == false -> is_available transition true-to-false in here triggers nop-recount and service stays "counted" - is that intentional?

connman/src/service.c
```diff
@@ -9794,6 +9853,8 @@ int __connman_service_init(void)
 
 	service_hash = g_hash_table_new_full(g_str_hash, g_str_equal,
 							NULL, service_removed);
+	service_type_hash = g_try_new0(GHashTable*,
```
spiiroin, line 9856: Why bother with try new if the result is never checked?

connman/src/service.c
```diff
@@ -510,10 +512,49 @@ static inline char *service_path(const char *ident)
...
+
+	DBG("%p/%s %s %c 1", service, service->name,
+					__connman_service_type2string(type),
+					add ? '+' : '-');
+
+	if (add) {
+		if (!is_available(service)) {
+			DBG("service %p/%s not available", service,
+							service->identifier);
+			return;
```
spiiroin, line 536: If a service is not going to be counted "in", should it be counted "out"? In if(add && available) insert(); else remove(); sort of way?

connman/src/service.c
LaakkonenJussi: If I remember correctly from my tests this actually is intentional as when this is triggered there is an update to the network few lines above (from `NULL` to a `ptr`) so it is assumed that condition changes if also the available value changes.

connman/src/service.c
LaakkonenJussi: The way different plugins seem to work: very differently they can try to add them in places before and after they are available. It is somewhat of a mess and this is just a guarantee to operate (hopefully) properly in all cases by just ignoring the addition in cases where it is called prior to being available.

Kind of difficult to think of a unit test for this either as the plugins just can work with different needs...

LaakkonenJussi: But this made me think if the `bool` is necessary after all. I tried this with `int` arrays first but with this hash table things are different..

LaakkonenJussi: Well, yeah, now I remember. VPN services and others alike that have different availability requirements do not work with plain availability check as they are always available, so removal does not work when relying on the availability. This is complex..

connman/src/service.c
LaakkonenJussi: Yep, only `g_new0` is to be used with small allocations, if it fails then it is deemed to fail as per to upstream coding conventions as well. Fixed with 97c3b643978731057463a21799e8bd362fb4a261

connman/src/service.c
LaakkonenJussi: Yep, that is better and I moved it to be done as first here to be on the safe side with 97c3b643978731057463a21799e8bd362fb4a261

connman/src/service.c
spiiroin: So, once a service has been counted as "added and available", it is ok for it to go back and forth between available and not-available without the count changing until service removal?

LaakkonenJussi: That does not happen, the `not available` is the case when loading the service, when a service has its network removed the count is decreased (any service with a changing network condition) or removed, which includes also the VPNs that are removed (has provider).

LaakkonenJussi: As it seemed that in some cases when connecting directly to a new network it can be in available state when it is first counted (or this is what I remember happening while testing) but in most of the cases it is not available when creating. Same function seems to get called twice in some cases, with same conditions except the network availability changes...

spiiroin: Ok. I have no idea what the pig picture is here and it just looked a bit suspicious. ..

spiiroin (approved):

rainemak:
> > Also @okodron review is appreciated.
> 
> CC @ikamaletdinov as well.

Ping @ikamaletdinov

ikamaletdinov:
Sorry for delay, LGTM!

I have just checked that with tons of networks around in our office. Seems that now we will have only 4 scans in 3 minutes with active display. So it is very good result. Not the ideal I guess but pretty cool! Before that fix we had hundreds of scans in 5 minute windows with same conditions.

PR #24, opened 2022-02-24
Check for pending requests also on hidden WiFi networks

LaakkonenJussi:
Hidden services do not keep the pending call when connecting as that is
forwarded for agent to handle, if present. This patch set adds simple search
to agent.c that can be used in service.c via agent-connman.c to check if the
hidden service has a pending connect request.

It was noticed that two consecutive connect requests for a hidden service was
possible and this resulted in adding two separate requests when agent was
present. First requests was placed as the pending request on the agent and the
second one was allowed as well but it was put into the queue. And when time-out
happened on the first one then the duplicate request would have been sent.

UI components relying on D-Bus API should not have to deal with this, it is
better to handle this case on ConnMan side.

### DIFF
--- a/connman/src/agent.c
+++ b/connman/src/agent.c
@@ -257,6 +257,34 @@ int connman_agent_queue_message(void *user_context,
 	return err;
 }
 
+bool connman_agent_queue_search(void *user_context, void *agent_data)
+{
+	struct connman_agent *agent = agent_data;
+	struct connman_agent_request *queue_data;
+	GList *iter;
+
+	DBG("user context %p agent %p", user_context, agent);
+
+	if (!agent || !user_context)
+		return false;
+
+	if (agent->pending && agent->pending->user_context == user_context) {
+		DBG("user context %p is pending for reply", user_context);
+		return true;
+	}
+
+	for (iter = agent->queue; iter; iter = iter->next) {
+		queue_data = iter->data;
+
+		if (queue_data && queue_data->user_context == user_context) {
+			DBG("user context %p is in queue", user_context);
+			return true;
+		}
+	}
+
+	return false;
+}
+
 static void set_default_agent(void)
 {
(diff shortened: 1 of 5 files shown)


Tomin1:
> [agent] Add search for pending requests. JB#57226

Should be 57266.

connman/src/agent.c
```diff
@@ -257,6 +257,32 @@ int connman_agent_queue_message(void *user_context,
...
+
+	DBG("user context %p agent %p", user_context, agent);
+
+	if (!agent || !user_context)
+		return false;
+
+	len = g_list_length(agent->queue);
+
+	for (i = 0; i < len; i++) {
+		queue_data = g_list_nth_data(agent->queue, i);
```
Tomin1, line 275: This is now O(N²) algorithm. Just thinking that to me it would make sense to iterate the list without using indices like below or use `g_list_find_custom()`.
```C
for (GList *elem = agent->queue; elem != NULL; elem = elem->next) {
    // check elem
}
```

LaakkonenJussi: As my eyes fooled me to see `GList` as `GSList`.

Tomin1: > I really did not yet think this but to get something for the test.

Ok. I didn't realize that myself.

LaakkonenJussi:
> > [agent] Add search for pending requests. JB#57226
> 
> Should be 57266.

Typo, fixd.

connman/src/agent.c
LaakkonenJussi: I think this is ok now.

LaakkonenJussi:
Maybe the check `__connman_agent_is_request_pending()` should be done always in addition to checking the `service->pending` by ignoring whether the service is hidden or not..

Tomin1 (approved):

PR #25, opened 2022-03-01
rtnl: Add ARPHRD_PUREIP netdev type.

bmigunov:
Since `ARPHRD_PUREIP` _netdev_ type has been introduced in kernel on some of
the MTK-based devices, there have been issues caused by the subject.

This _netdev_ type needs to be handled by `ipconfig` module, so this commit
adds the support of the `ARPHRD_PUREIP` by handling appropriate _ntelink_
events.


connman/src/rtnl.c
```diff
@@ -51,6 +51,10 @@
 #define ARPHRD_RAWIP_1 (519) /* Vanilla kernel */
 #define ARPHRD_RAWIP_2 (530) /* Some android kernels */
 
+/* ARPHRD_PUREIP has been introduced in some MTK kernels (in intention to
+ * get rid of link local address creation for the ccmni net devices) */
```
LaakkonenJussi, line 55: Untrue, well, maybe for MTK but in general reason was this https://lkml.org/lkml/2021/6/24/76 and the solution is not really accepted into kernels as of https://lkml.kernel.org/netdev/YNmWwSsZ02iigiHC@kroah.com/

Not yet sure if this is to be included. Upstream might criticize this for sure.

And, multi line comments are formatted (coding-style.txt M2):
```
/*
 * Line1.
 * Line2.
 */
```

bmigunov: Comment section fixed.
It is possible for me to roll current MTK kernel's `ccmni` interface _netdev_ type back to `ARPHRD_PPP` as long as it will not lead to any kind of issues (presumably).
But still, I'll leave this PR as a suggestion because there are no guarantees that `ARPHRD_PUREIP` will not be used in the future MTK projects.

PR #26, opened 2022-04-04
Move resolv.conf to /run/connman/resolv.conf

daniellinjama:
Move resolv.conf to /run/connman/resolv.conf to make resolver configuration available inside sailjail

### DIFF
--- a/connman/Makefile.am
+++ b/connman/Makefile.am
@@ -255,6 +255,7 @@ AM_CFLAGS = @DBUS_CFLAGS@ @GLIB_CFLAGS@ @XTABLES_CFLAGS@ \
 				-DPLUGINDIR=\""$(build_plugindir)"\" \
 				-DSCRIPTDIR=\""$(build_scriptdir)"\" \
 				-DDEFAULT_STORAGE_ROOT=\""$(storageroot)\"" \
+				-DRUNSTATEDIR=\""$(runstatedir)"\" \
 				-DCONFIGDIR=\""$(configdir)\""
 
 if VPN
@@ -272,6 +273,7 @@ src_connmand_CFLAGS = @DBUS_CFLAGS@ @GIO_CFLAGS@ @GOBJECT_CFLAGS@ @GLIB_CFLAGS@ 
 				-DSCRIPTDIR=\""$(build_scriptdir)"\" \
 				-DDEFAULT_STORAGE_ROOT=\""$(storageroot)"\" \
 				-DDEFAULT_USER_STORAGE=\""$(userstorage)"\" \
+				-DRUNSTATEDIR=\""$(runstatedir)"\" \
 				-DCONFIGDIR=\""$(configdir)"\" \
 				-DDEFAULT_VPN_STATEDIR=\""$(vpn_statedir)"\" \
 				-I$(builddir)/src
@@ -289,6 +291,7 @@ vpn_connman_vpnd_CFLAGS = @DBUS_CFLAGS@ @GLIB_CFLAGS@ @DBUSACCESS_CFLAGS@ \
 				-DSCRIPTDIR=\""$(build_scriptdir)"\" \
 				-DDEFAULT_STORAGE_ROOT=\""$(storageroot)"\" \
 				-DDEFAULT_USER_STORAGE=\""$(userstorage)"\" \
+				-DRUNSTATEDIR=\""$(runstatedir)"\" \
 				-DCONFIGDIR=\""$(configdir)"\" \
 				-I$(builddir)/vpn
 
--- a/rpm/connman.spec
+++ b/rpm/connman.spec
@@ -198,7 +198,8 @@ This package provides OpenFortiNet VPN plugin for connman.
     --disable-wispr \
     --with-systemdunitdir=%{_unitdir} \
     --enable-systemd \
-    --with-tmpfilesdir=%{_prefix}/lib/tmpfiles.d
+    --with-tmpfilesdir=%{_prefix}/lib/tmpfiles.d \
+    runstatedir=/run
 
 %make_build
 
@@ -235,7 +236,7 @@ fi
 
 %post
 # These should match connman_resolvconf.conf rules
-%define connman_run_dir /var/run/connman
+%define connman_run_dir /run/connman
 %define run_resolv_conf %{connman_run_dir}/resolv.conf
 %define etc_resolv_conf %{_sysconfdir}/resolv.conf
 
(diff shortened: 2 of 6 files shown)


daniellinjama:
Issue reported on Sailfish OS Forum https://forum.sailfishos.org/t/browser-not-working-when-opened-from-app-drawer/10829

rpm/connman.spec
```diff
@@ -198,7 +198,8 @@ This package provides OpenFortiNet VPN plugin for connman.
     --disable-wispr \
     --with-systemdunitdir=%{_unitdir} \
     --enable-systemd \
-    --with-tmpfilesdir=%{_prefix}/lib/tmpfiles.d
+    --with-tmpfilesdir=%{_prefix}/lib/tmpfiles.d \
+    runstatedir=/run
```
spiiroin, line 202: Does this "runstatedir=/run" actually work? [I was thinking that this ought to be handled via "--runstatedir=xxx", but the configure script that is generated does not seem to support such option so idk]

AFAICT for resolv conf symlink it would suffice that %post script and tmpfiles config change from /var/run to /run, but ... There are bunch of hard coded "/var/run" paths here and there in the code base. Might be nice to have those to in sync, or?

[for reference: a quick and dirty search and replace hack I made earlier can be found from here:
https://github.com/spiiroin/connman/tree/jb57910_legacy_var_run_wip ]

LaakkonenJussi: I think support for changing the `run` dir must be added to `configure.ac` and use must be extended to where the path is still hardcoded. And the value comes from m4 automake I think, see https://www.mail-archive.com/connman@connman.net/msg18690.html 

Upstream was happy with using `/var/run` still. Maybe an upstream patch is also in place.

LaakkonenJussi: The upstream commits:
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=0db8be18510e529a33b15735b339bd42dbd39c12

https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=c1211fbe653590c017d1f3b992a94e5b8bb22cdc

daniellinjama: > Does this "runstatedir=/run" actually work? [I was thinking that this ought to be handled via "--runstatedir=xxx", but the configure script that is generated does not seem to support such option so idk]
> 
> AFAICT for resolv conf symlink it would suffice that %post script and tmpfiles config change from /var/run to /run, but ... There are bunch of hard coded "/var/run" paths here and there in the code base. Might be nice to have those to in sync, or?
> 
> [for reference: a quick and dirty search and replace hack I made earlier can be found from here: https://github.com/spiiroin/connman/tree/jb57910_legacy_var_run_wip ]

Yes, it works. Option `--runstatedir` is supported only with autoconf 2.70 or greater. Without `runstatedir=/run` generated Makefile contains `runstatedir = ${localstatedir}/run` and with `runstatedir=/run` generated `runstatedir = /run`. I found this workaround from [Kea documentation](https://kea.readthedocs.io/en/kea-1.6.2/arm/install.html).

I'm not sure if hard coded `/var/run` paths should be replaced but at least resolver configuration is working correctly with sailjailed applications

daniellinjama: > I think support for changing the `run` dir must be added to `configure.ac` and use must be extended to where the path is still hardcoded. And the value comes from m4 automake I think, see https://www.mail-archive.com/connman@connman.net/msg18690.html
> 
> Upstream was happy with using `/var/run` still. Maybe an upstream patch is also in place.

Extended hard coded `/var/run` paths with `RUNSTATEDIR`. Created another branch [fix-sailjail-resolv-autoconf270](https://github.com/daniellinjama/connman/tree/fix-sailjail-resolv-autoconf270) for `--runstatedir=/run` variant but autoconf 2.70 is not available on Sailfish SDK

LaakkonenJussi: Looks fine now, and builds ok with or without the change in the `.spec`. Regarding upstream, this is something we'd like to keep in sync so some of this is to be upstreamed. One commit is hard to split for that purpose. Therefore, I'd propose splitting this into smaller commits, for example (commit texts are just suggestion, it is good to have some additional explanatory commit message especially for upstream):

- rpm/connman.spec is our fork only change, not upstreamed
  - `[packaging] Replace legacy /var/run with /var.`

- test-vpn-util.c is only in our fork, not upstreamed (upstream does not have much unit tests..)
  -  `[unit] Use RUNSTATEDIR in vpn util tests.`

- Makefile.{am,plugins} would be upstreamable
  -  `build: Support configurable run dir with RUNSTATEDIR` 
 
- vpnc.c and vpn-util.c would be upstreamable, could be combined even
  -  `vpn: Replace hardcoded paths with RUNSTATEDIR`

If you need help in making upstream patches I can write instructions or do it for you if you wish.

spiiroin: > `runstatedir=/run` ... I found this workaround from [Kea documentation](https://kea.readthedocs.io/en/kea-1.6.2/arm/install.html).
Thanks for this. All I found was problem statements/bugs, so nice to know that there is a workaround too.

LaakkonenJussi (commented):
LGTM, and as you sent the upstream patches, well, it does not matter if we take this as a big commit. I'll add the annotated tag after merge to contain necessary bits for the build system.

LaakkonenJussi (approved):

spiiroin (approved):
Couple of cosmetic nits, but: builds, installs, and rectifies the symlink

connman/Makefile.am
```diff
@@ -255,7 +255,8 @@ AM_CFLAGS = @DBUS_CFLAGS@ @GLIB_CFLAGS@ @XTABLES_CFLAGS@ \
 				-DPLUGINDIR=\""$(build_plugindir)"\" \
 				-DSCRIPTDIR=\""$(build_scriptdir)"\" \
 				-DDEFAULT_STORAGE_ROOT=\""$(storageroot)\"" \
-				-DCONFIGDIR=\""$(configdir)\""
+				-DCONFIGDIR=\""$(configdir)\"" \
+				-DRUNSTATEDIR=\""$(runstatedir)\""
```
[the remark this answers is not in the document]
daniellinjama: I used same syntax as configdir but for some reason escaped quotesa are different on AM_CFLAGS. Reordered defines to match with upstream patch and changed escaped quotes, but now this line differs from upstream patch

rpm/connman.spec
daniellinjama: > Cosmetic, but as there is variable defined for /run, this could use "runstatedir=%{connman_run_dir}" too?

runstatedir is "/run" and connman_run_dir is "/run/connman"

spiiroin: > runstatedir is "/run" and connman_run_dir is "/run/connman"
mm, yes. Apparently I got confused.

Thaodan: Should it maybe match the other options syntax style? All others use --with-foobar=bar.

LaakkonenJussi: That comes from autoconf, with the older version we have runstatedir is to be used. Later, with autoconf 2.70 `--runstatedir` is to be used. It seems kind of unnecessary to make an explicit support for that now.

PR #27, opened 2022-04-05
Change (WiFi) regdomain when timezone changes

LaakkonenJussi:
Add an optional feature to change the regdomain when timezone is changed. This is done by reading the ISO3166 country code from `/usr/share/zoneinfo/zone1970.tab` using the set timezone as key, the first code is used as the alpha2 code.

Also add config options for `Localtime` and to enable this feature with `RegdomFollowsTimezone`.


connman/src/main.c
```diff
@@ -118,6 +120,7 @@ static struct {
 	.vendor_class_id = NULL,
 	.fallback_device_types = NULL,
 	.enable_login_manager = false,
+	.localtime = DEFAULT_LOCALTIME,
```
inzanity, line 124: Note: this assigns a `const char *` to a `char *`.

connman/src/main.c
```diff
@@ -544,14 +549,25 @@ static void parse_config(GKeyFile *config)
...
 	if (!error)
 		connman_settings.enable_login_manager = boolean;
 
-	g_strfreev(str_list);
+	g_clear_error(&error);
+
+	str = __connman_config_get_string(config, group, CONF_LOCALTIME,
+				&error);
+	if (!error) {
+		connman_settings.localtime = str;
```
inzanity, line 570: This is not `free()`d on shutdown.

inzanity (approved):
In general looks reasonable, couple of minor remarks.

connman/src/main.c
LaakkonenJussi: This should take care of it 481c63fe79c452579491f83ab27753ac3f6d4224

connman/src/main.c
LaakkonenJussi: Yep, modified the use of this to be in line with the status check urls with 481c63fe79c452579491f83ab27753ac3f6d4224 so the default value is returned in the getter when `localtime` is `NULL`.

LaakkonenJussi:
One question still remains that when this feature is enabled to change regdom based on timezone it should be disabled on ofono plugin so only either of them changes the regdom. After all when Ofono changes timezone based on the automatic setting the change would then drain to ConnMan as well with this change.

connman/plugins/ofono.c
```diff
@@ -1680,6 +1681,10 @@ static void netreg_update_regdom(struct modem_data *modem,
 	char *alpha2;
 	int mcc;
 
+	/* Do not change regdom here if it is set to follow timezone. */
+	if (connman_setting_get_bool("RegdomFollowsTimezone"))
```
inzanity, line 1685: Would it make sense to have a method to use SIM if possible, and timezone only as a backup?

LaakkonenJussi: If timezone is set to automatic and updated via another service, or set via ConnMan even it should be the cellular stack changing it -> the change should be noted when RegdomFollowsTimezone is set. I was just thinking if it can create some corner cases if used only as a backup in situations where Ofono plugin sets the regdom to X and for some reason user has selected a certain timezone, which should be preferred as in that case it is unknown which option the user wishes the system to follow and this gives the power to do so.

inzanity: At least I (as a user) wouldn't expect the choice of timezone to affect the wifi regdom.

LaakkonenJussi: This is just the case, relying on timezone would be a kind of solution for every possible case. Having cellular support and a SIM present is a somewhat special case ;).

But I think many do not even expect that the cellular stack changes the regdom? 

To facilitate this fully to have timezone as a backup would need quite a lot of new code I think. `technology.c` has no way of detecting who has changed the value. But would you suggest that the timezone at startup defines the base regdom and then that is changed by Ofono plugin? Eventually the timezone after all changes, or should change, according to the cellular network when it is supported -> both produce the same result. 

My argument here is that now with this feature user can keep a timezone and regdom.

monich: IMO network registration should take precedence. When available, it's a far more accurate guess than whatever the user has selected in the settings.

connman/src/timezone.c
```diff
@@ -226,18 +226,99 @@ static char *find_origin(void *src_map, struct stat *src_st,
...
+			/*
+			 * Multiple country codes can be listed, use the first
+			 * 2 chars.
+			 */
+			alpha2 = g_strndup(g_strstrip(tokens[0]), 2);
+			if (strlen(alpha2) != 2) {
+				connman_warn("Invalid ISO3166 code %s", alpha2);
+				g_free(alpha2);
+				alpha2 = NULL;
+				break;
```
spiiroin, line 289: 'break' skips freeing 'tokens' and 'line', or?

also, is there any chance that following input lines could contain valid entry i.e. is breaking out ok?

connman/src/timezone.c
```diff
@@ -226,18 +226,99 @@ static char *find_origin(void *src_map, struct stat *src_st,
...
+		return NULL;
+	}
+
+	DBG("read %s for %s", USR_SHARE_ZONEINFO_MAP, zone);
+	g_io_channel_set_encoding(channel, "UTF-8", NULL);
+
+	while (g_io_channel_read_line(channel, &line, &len, NULL, NULL) ==
+							G_IO_STATUS_NORMAL) {
+		if (!line || !*line || *line == '#' || *line == '\n')
+			continue;
```
spiiroin, line 269: non-null 'line' leaked by continue?

connman/src/timezone.c
```diff
@@ -226,18 +226,99 @@ static char *find_origin(void *src_map, struct stat *src_st,
...
+
+		/* File format: Countrycodes Coordinates TZ Comments */
+		tokens = g_strsplit_set(line, " \t", 4);
+		if (!tokens) {
+			connman_warn("line %s failed to parse", line);
+			continue;
```
spiiroin, line 275: 'line' leaked?

connman/src/timezone.c
```diff
@@ -226,18 +226,99 @@ static char *find_origin(void *src_map, struct stat *src_st,
...
+		}
+
+		g_strfreev(tokens);
+		g_free(line);
+
+		if (alpha2)
+			break;
+	}
+
+	g_io_channel_shutdown(channel, FALSE, NULL);
```
spiiroin, line 302: doing a shutdown in this context (=reading from file) feel a bit pointless.
also there is no g_io_channel_unref() so 'channel' is leaked

connman/src/timezone.c
```diff
@@ -226,18 +226,99 @@ static char *find_origin(void *src_map, struct stat *src_st,
...
+		return NULL;
+
+	fd = open(USR_SHARE_ZONEINFO_MAP, O_RDONLY | O_CLOEXEC);
+	if (fd < 0) {
+		connman_warn("failed to open zoneinfo map %s",
+							USR_SHARE_ZONEINFO_MAP);
+		return NULL;
+	}
+
+	if (fstat(fd, &st) < 0) {
```
spiiroin, line 249: given that 'st' is not used, what is the point with this fstat()?

connman/src/timezone.c
```diff
@@ -336,7 +426,7 @@ int __connman_timezone_change(const char *zone)
 		return -EIO;
 	}
 
-	err = write_file(map, &st, ETC_LOCALTIME);
+	err = write_file(map, &st, connman_setting_get_string("Localtime"));
```
spiiroin, line 429: What is going on in here?
To me it looks like it reads data from /usr/share/zoneinfo/xxx and then writes it to where ever /etc/localtime or similar symlink happens to point to i.e. potentially corrupts some random zoneinfo file?

LaakkonenJussi: This is part of the [clock.c D-Bus API that can change timezone](https://github.com/sailfishos/connman/blob/master/connman/src/clock.c#L288). Maybe we'd at least need a config option to disable that `__connman_timezone_change()`?

connman/src/timezone.c
LaakkonenJussi: Restructured 43d24dac8eeb7b1a50f19f9a4517c84c49cddae8 to address this

connman/src/timezone.c
LaakkonenJussi: Restructured 43d24dac8eeb7b1a50f19f9a4517c84c49cddae8 to address this

connman/src/timezone.c
LaakkonenJussi: Fixed with 43d24dac8eeb7b1a50f19f9a4517c84c49cddae8

connman/src/timezone.c
LaakkonenJussi: OK, if it is not really needed I replaced shutdown with just unref with 43d24dac8eeb7b1a50f19f9a4517c84c49cddae8

connman/src/timezone.c
spiiroin: Hmm, on closer look: if destination path is a symlink, write_file() makes an attempt to remove it before writing. So, it should not corrupt files possible symlinks point to. But it still would break the symlink chain for which we have built-in expectations here and there - mainly that tail end of realpath of /etc/localtime = name of the timezone, possibly time/date settings & timed itself.

Having an option to disable this op from dbus api might be nice. But as this is something that already has been there for ages, it could be a separate task too.

LaakkonenJussi: Yes, it works exactly like that. Only think I was also worried was that it will break our timed set localtime. We can surely do that later as this D-Bus call is not apparently used as of now.

connman/src/timezone.c
LaakkonenJussi: I think I just simply forgot to do this db56d888e1f90585888398b499571399d4ceb249

spiiroin: so, regular files and symlinks to regular files are ok, but not sockets, pipes, etc - makes sense

spiiroin (approved):
All my concerns have been addressed

LaakkonenJussi:
Well, I'll merge this to see how it works in devel. If we need to further make changes, just make new tasks. But I think this is quite consistent way of doing things. And configurable.

PR #28, opened 2022-04-07
[connman] Apply upstream rtnl patches, revert debug disable commit. JB#57933

LaakkonenJussi:
Applied a patch that fixes the alignment of netlink messages. And also couple of others non-related to maintain consistency in rtnl.c.

Reverted old debug disable commit that was not just proper. `DBG()`s are handled via bootup option or D-Bus toggle with dbuslog-tools.

### DIFF
--- a/connman/src/detect.c
+++ b/connman/src/detect.c
@@ -49,7 +49,7 @@ static void detect_newlink(unsigned short type, int index,
 	struct connman_device *device;
 	enum connman_device_type devtype;
 
-//	DBG("type %d index %d", type, index);
+	DBG("type %d index %d", type, index);
 
 	devtype = __connman_rtnl_get_device_type(index);
 
--- a/connman/src/device.c
+++ b/connman/src/device.c
@@ -1316,7 +1316,7 @@ struct connman_device *connman_device_create_from_index(int index)
 		return NULL;
 
 	if (__connman_device_isfiltered(devname)) {
-//		DBG("Ignoring interface %s (filtered)", devname);
+		DBG("Ignoring interface %s (filtered)", devname);
 		g_free(devname);
 		return NULL;
 	}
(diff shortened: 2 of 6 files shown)


connman/src/detect.c
```diff
@@ -27,7 +27,13 @@
 
 #include "connman.h"
 
+struct device_details {
+	enum connman_device_type type;
+	int index;
+};
+
 static GSList *device_list = NULL;
+static gint delayed_newlink_ids[MAX_CONNMAN_SERVICE_TYPES] = { 0 };
```
inzanity, line 36: Can we relay on there only being one new link per service type?

LaakkonenJussi: Yes but this is more of an idea level approach still that might this be the cause even. Which is why I just added commit to add delayed only when there is no MAC. Didn't want to overdo this yet.

LaakkonenJussi: Yep, this made things worse when there is no MAC set for a device. Removed.

LaakkonenJussi:
> @inzanity Added few untested commits, which are more of a workaround for the issue when there are systems that do report the netlink message BEFORE there is a MAC set, I've seen this (as a result MAC is 000000000000):
> 
> ```
> connmand[615]: eth0 no DEVTYPE, defaulting to ethernet
> kernel: rk_gmac-dwmac fe300000.ethernet: rk_get_eth_addr: mac address: xx:xx:xx:xx:xx:xx
> kernel: eth0: device MAC address xx:xx:xx:xx:xx:xx
> ```
> 
> And it can happen in boot, not 100% sure but this would avoid any odd races with kernel regarding this. It would be better to fix the kernels I guess but...
> 
> Haven't tested that yet at all. Too late, Friday.

Removed, only the upstream patches remain and reverting of old commit.

PR #29, opened 2022-05-12
WIP: Enforce removal of additional mobile data default routes

LaakkonenJussi:


connman/plugins/sailfish_ofono.c
```diff
@@ -137,6 +138,121 @@ struct plugin_data {
...
+	int err;
+
+	/* Delete all gateways that are not for the current interface */
+	if (item->index == index)
+		return;
+
+	DBG("deleting %d %s %s %d", item->index, item->dst, item->gw,
+								item->metric);
+
+	/* Interested only in default gw routes that have no prefix */
```
LaakkonenJussi, line 238: Just a reminder for self `struct in6_rtmsg` does contain the prefix length as well.

LaakkonenJussi:
If this should be always active I think this ofono bit could be even a separate plugin, to always block those default routes set via RA for the additional `rmnet_data` interface. Then simply monitoring services certain required actions could be made.

abranson (approved):
Fixes my mobile ipv6 connectivity routing problems!

connman/plugins/sailfish_ofono.c
LaakkonenJussi: This was available otherwise.

abranson:
But now it doesn't anymore. I think the move to the plugin broke something.

LaakkonenJussi:
> But now it doesn't anymore. I think the move to the plugin broke something.

I think this is now fixed. And it would be good to get some eyes on the code as well.

connman/include/rtnl.h
```diff
@@ -47,17 +49,25 @@ void connman_rtnl_remove_watch(unsigned int id);
 struct connman_rtnl {
 	const char *name;
 	int priority;
+	bool handle_rtprot_ra;
```
monich, line 52: If this is a public API exposed to external plugins then sticking new fields into the middle of the structure doesn't look like a good idea. On the other hand, connman doesn't seem to care much about stability of the plugin interface, this change doesn't make much of a difference, feel free to ignore.

connman/include/inet.h
```diff
@@ -56,11 +56,12 @@ int connman_inet_set_ipv6_address(int index,
 int connman_inet_clear_ipv6_address(int index,
 					struct connman_ipaddress *ipaddress);
 int connman_inet_add_ipv6_network_route(int index, const char *host,
-					const char *gateway, unsigned char prefix_len);
+					const char *gateway,
+					unsigned char prefix_len, short metric);
```
monich, line 60: Again, if this is a public interface to the external plugins (looks like it - these functions are exported from connman and the header is a part of connman-devel package), function signatures must not change over time.

connman/plugins/blacklist_monitor.c
```diff
@@ -0,0 +1,610 @@
...
+
+static GSList *rtnl_gw_list = NULL;
+static GSList *rtnl_gw_restore_list = NULL;
+
+static struct rtnl_gw *new_rtnl_gw(int index, const char *gw, int metric)
+{
+	struct rtnl_gw *item;
+
+	item = g_try_new0(struct rtnl_gw, 1);
+	if (!item) {
```
monich, line 50: This `g_try_new0` and the subsequent check of the return value doesn't make sense because e.g. `g_strdup()` below does `g_new` and doesn't check the return value. OOM condition is unrecoverable (because the checks are random and because of over-commit).

connman/plugins/blacklist_monitor.c
```diff
@@ -0,0 +1,610 @@
...
+
+	if (route_a->index == route_b->index &&
+				!g_strcmp0(route_a->gw, route_b->gw) &&
+				route_a->metric == route_b->metric)
+		return 0;
+
+	return -1;
+}
+
+static gint compare_route_by_index(gconstpointer a, gconstpointer b) {
```
monich, line 95: Inconsistent use of `{`

connman/include/rtnl.h
monich: And there's also global variable called `handle_rtprot_ra` in `src/rtnl.c` - what's the relationship between those?

connman/src/rtnl.c
```diff
@@ -1126,35 +1154,66 @@ static bool is_route_rtmsg(struct rtmsg *msg)
 		return false;
 
 	if (msg->rtm_protocol != RTPROT_BOOT &&
-			msg->rtm_protocol != RTPROT_KERNEL)
+					msg->rtm_protocol != RTPROT_KERNEL) {
+
+		if (handle_rtprot_ra && msg->rtm_protocol == RTPROT_RA)
```
monich, line 1159: Please double-check use of `rtnl->handle_rtprot_ra` vs global `handle_rtprot_ra`. It wasn't obvious to me.

connman/include/inet.h
spiiroin: afaict previously fixed metric=1 value waws . so perhaps api compatibility could be retained by adding a new function xxx_with_metric() and making existing connman_inet_add_ipv6_network_route() call it with metric value 1?

spiiroin (approved):
Did not spot any showstoppers. Left some remarks, check if there is something that needs to be addressed in there.

connman/plugins/blacklist_monitor.c
```diff
@@ -0,0 +1,610 @@
...
+{
+	const struct rtnl_gw *route_a = a;
+	const struct rtnl_gw *route_b = b;
+
+	if (route_a->index == route_b->index &&
+				!g_strcmp0(route_a->gw, route_b->gw) &&
+				route_a->metric == route_b->metric)
+		return 0;
+
+	return -1;
```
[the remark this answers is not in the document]
LaakkonenJussi: Cleaned up these for consistency a bit: 3e7c78cfe09a61704f71326ad4192a2e86d9451c so in the future it does not hopefully confuse anymore?

connman/plugins/blacklist_monitor.c
LaakkonenJussi: Yeah, fixed along 3e7c78cfe09a61704f71326ad4192a2e86d9451c

connman/plugins/blacklist_monitor.c
```diff
@@ -0,0 +1,610 @@
...
+	service_index = connman_network_get_index(network);
+	if (service_index != index)
+		remove_route(item, GINT_TO_POINTER(service_index));
+	else /* do nothing for connected gw */
+		DBG("index %d is the default service -> not adding", index);
+
+	goto out;
+
+add:
+	free_item = add_gateway_to_rtnl_list(item) != 0 ? true : false;
```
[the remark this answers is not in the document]
LaakkonenJussi: Setting `NULL` is a good idea, I like it, clears the code, thanks. Changed in 6a6cd3e0d54063f0ee355c5c8ab76eff91b61895

connman/plugins/blacklist_monitor.c
```diff
@@ -0,0 +1,610 @@
...
+		return list;
+	}
+
+	*err = -ENOENT;
+
+	for (iter = list; iter; iter = iter->next) {
+		route = iter->data;
+
+		if(!compare_route(route, item)) {
+			list = g_slist_remove_link(list, iter);
```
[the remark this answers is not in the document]
LaakkonenJussi: Yeah, I think this approach is from some old attempt. That is cleaner, changed in 6a6cd3e0d54063f0ee355c5c8ab76eff91b61895

connman/plugins/blacklist_monitor.c
LaakkonenJussi: I forgot that small allocations are let to fail as that is a bigger issue anyways when those fail. Changed in 6a6cd3e0d54063f0ee355c5c8ab76eff91b61895 just to use `g_new0` and ignore error, if if crashes it is not the only thing that is bound to crash in such case.

connman/src/rtnl.c
```diff
@@ -1126,35 +1154,66 @@ static bool is_route_rtmsg(struct rtmsg *msg)
...
+			return true;
```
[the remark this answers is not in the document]
LaakkonenJussi: This just follows the style of the function. Nothing more.

connman/include/inet.h
LaakkonenJussi: Changed to add a new function there, so everything uses the `_with_metric()` keeping all other remaining function uses the same: 6042574547fa41cee13f3715b5575c323a3d4e48 + 30cf0fa36c0fa42d3befc25eaa09c21fde965ac6 + 6d1254074d63993fd705d0a4937f5047b5fdc06c + 37018906609f032c16734686a054efb2b592c465

connman/include/rtnl.h
LaakkonenJussi: In normal use, without the plugin those IPv6 `RTPROT_RA` messages should not be processed. The `rtnl.c` option indicates simply that those kind of messages are required to be processed, usually the RA added content is left for kernel to handle but with these blacklisted interfaces we need to see also those changes. 

I'm not really comfortable with the idea of breaking the API either but each plugin registering for IPv6 route changes may not want to handle the RA generated netlink messages. This is really a special case when the blacklisted interfaces cause issues. Other way would be that the message type is included in the callback function, which would not break the API. I'll see if that is cleaner.

connman/src/rtnl.c
LaakkonenJussi: The changes 324012e324bcea0cab51a5758a84e9d242a5ec94 + a46ed14c2d6c6a596f03c62602ea0bb5d5d1ad20 + e07773823fa3b1527d12f00781339e61165028a9 make this go the other way. And as I suggested in https://github.com/sailfishos/connman/pull/29#discussion_r909543189 it might be better + more generic way to add a functionality to support the varying kernel netlink user type protocols to be handled.

connman/src/rtnl.c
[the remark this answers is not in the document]
LaakkonenJussi: This will change later when the user rtprotocol type handling is implemented. No use touching this anymore. Setting as resolved.

Tomin1 (approved):

connman/include/rtnl.h
LaakkonenJussi: Dealt with as little as possible damage to the API. Marking resolved.

connman/plugins/blacklist_monitor.c
```diff
@@ -0,0 +1,610 @@
...
+	if (!item) {
+		*err = -EINVAL;
+		return list;
+	}
+
+	DBG("index %d dst %s gw %s metric %d", item->index, IPV6_ANY, item->gw,
+								item->metric);
+
+	if (!list) {
+		*err = 0;
```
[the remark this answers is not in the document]
LaakkonenJussi: Eventually this should not happen but... I guess not a biggie. Resolving.

connman/plugins/blacklist_monitor.c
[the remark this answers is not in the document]
LaakkonenJussi: Squashed already, I guess this is ok, resolving.

PR #31, opened 2022-07-28
Revert "[wifi] Implement a heurestic for determening a weak WiFi."

Mukhamedgarif:
This heuristic causes errors with eap wifi networks in areas with many wifi networks. 
It seems that heuristic drops performance because of frequent calls of wifi_bss_weak function.
This reverts commit e7b19da35f5de5b3d6a88b7dfe85a96397d0c95a.

### DIFF
--- a/connman/plugins/sailfish_wifi.c
+++ b/connman/plugins/sailfish_wifi.c
@@ -65,11 +65,7 @@
 #define WIFI_BSS_COUNT_SMALL (10)
 #define WIFI_BSS_COUNT_MAX (100)
 #define WIFI_WEAK_RSSI (-85) /* dBm */
-#define WIFI_WEAK_RSSI_MAX (-65) /* dBm */
 #define WIFI_WEAK_BSS_MIN_SIGHTINGS (3)
-
-#define WIFI_SERVICE_COUNT_INTERVAL (20)
-#define WIFI_RSSI_INCREASE_INTERVAL (5)
 
 /* Access point (tethering) configuration */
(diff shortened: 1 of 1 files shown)


attah:
What does the authentication mode have to do with this signal level heuristic?

Mukhamedgarif:
When you connects to wpa2-eap network it takes two attempts to establish connection.
According to debug logs wifi_bss_weak is called very often. I have 40 networks in view, which makes things worse.  
Some speed hacks  fixed it. For example, I've replaced division by 20 with shift left by 4 and it works.

attah:
Thanks for the explanation. Very strange/interesting that such a minor optimization was enough.

Mukhamedgarif:
I think current performance is near the edge, so minor optimization helped. But there is no guarantee that for 100 wifi networks it will work as well. As I know division is performance expencive operation, so it was the first thing to change.

LaakkonenJussi:
> When you connects to wpa2-eap network it takes two attempts to establish connection. According to debug logs wifi_bss_weak is called very often. I have 40 networks in view, which makes things worse. Some speed hacks fixed it. For example, I've replaced division by 20 with shift left by 4 and it works.

So is the real issue that the network you are connecting to may be removed by this heurestic? Then it is just requiring a change to keep the networks that are being connected to. By reverting a single commit from the larger change is not really something that should be done here as it leaves the rest kind of crippled/useless.

pvuorela:
No activity. We shouldn't revert the change.

PR #35, opened 2022-12-09
[connman] Apply upstream fix for HTTP response crash. Fixes JB#59564

LaakkonenJussi:
Applies an upstream patch: "gweb: Fix crash on malformed http response" 
[22901212105055b24f73504a74f5a57eee809777](https://git.kernel.org/pub/scm/network/connman/connman.git/commit/gweb/gweb.c?id=22901212105055b24f73504a74f5a57eee809777)

to address issue caused by a CVE-fix "gweb: Fix OOB write in received_data()"
[d1a5ede5d255bde8ef707f8441b997563b9312bd](https://git.kernel.org/pub/scm/network/connman/connman.git/commit/gweb/gweb.c?id=d1a5ede5d255bde8ef707f8441b997563b9312bd).

Also drop the use of errors for reading IPv4 status url from the configuration file. With
IPv6 this was also ignored and the default is being returned when the status url is not
set. This caused a Glib warning as GError was not cleared.

### DIFF
--- a/connman/src/main.c
+++ b/connman/src/main.c
@@ -518,7 +518,7 @@ static void parse_config(GKeyFile *config)
 	parse_perm(config, group, CONF_UMASK, &connman_settings.umask);
 
 	connman_settings.ipv4_status_url = __connman_config_get_string(config,
-					group, CONF_STATUS_URL_IPV4, &error);
+					group, CONF_STATUS_URL_IPV4, NULL);
 
 	connman_settings.ipv6_status_url = __connman_config_get_string(config,
 					group, CONF_STATUS_URL_IPV6, NULL);
--- a/connman/gweb/gweb.c
+++ b/connman/gweb/gweb.c
@@ -809,6 +809,9 @@ static void handle_multi_line(struct web_session *session)
 	char *str;
 	gchar *value;
 
+	if (!session->result.last_key)
+		return;
+
 	str = session->current_header->str;
 
 	if (str[0] != ' ' && str[0] != '\t')


Thaodan (approved):
LGTM

connman/src/main.c
```diff
@@ -532,6 +532,8 @@ static void parse_config(GKeyFile *config)
 		g_free(str);
 	}
 
+	g_clear_error(&error);
```
spiiroin, line 535: 1st thought: why not put this closer to where the error can end up being set?

2nd thought: after locating ^ place ->  the error value from ipv4 lookup is not used, and similar ipv6 lookup passes null error (in fact almost all lookups are like this) -> why not do that for ipv4 lookup too and thus have no dontcare error to clear here?

martyone: > the error value from ipv4 lookup is not used, and similar ipv6 lookup passes null error...

Yup, it seems it just has been forgotten here in the ipv4 case https://github.com/sailfishos/connman/commit/519ced062e926f6d1ebd5adffa8c27c0fa8a52da#diff-56a8ba3b41201b3671fbec69b15cc878e34f61922921e4f821d79a7bf53c1c96R403

LaakkonenJussi: Yeah, as the aim is to just use the default if the configuration is missing it might be best to drop the error from the IPv4 case as well. I'll force push the change.

LaakkonenJussi: Done, dropped from IPv4 with 8af344097b142cacf064ead18dcebc8923ef6a20. Resolving this.

martyone (approved):
I would just suggest to 1) add the common "(cherry picked from commit ...)" note to the first commit and 2) consider to tidy up the changelog messages (the one in PR covers both commits while the second commit comes with its own message too. So maybe use own message for each commit?)

LaakkonenJussi:
> I would just suggest to 1) add the common "(cherry picked from commit ...)" note to the first commit and 2) consider to tidy up the changelog messages (the one in PR covers both commits while the second commit comes with its own message too. So maybe use own message for each commit?)

Thanks. I don't usually edit the upstream commit texts, and not the commits themselves if they apply without conflicts. The logic here is to keep the tracing of the commits easier, I usually put commits that can be upstreamed with both formats:
 - `source: commit text`
 - `[source] commit text. JB#XXXXX`

So from history it is easy to detect that if the change is fork only then it has []-tag and the change in case of upgrade (hopefully someday...) needs to be kept and changing that part might be bad for us.

I've been having these summary texts in the PR titles as that is aiming to give a good overall view on what was done in that particular PR. Then the commit messages can detail the specific changes.

martyone:
> The logic here is to keep the tracing of the commits easier

The easiest to trace are the commits with "(cherry picked from commit ...)" note as the `-x` option to `git cherry-pick` does :)

> I've been having these summary texts in the PR titles as that is aiming to give a good overall view on what was done in that particular PR. Then the commit messages can detail the specific changes.

Yeah, summary in PR message is great, but keep in mind that if you write it as a changelog item, then it will appear in changelog on the same level as the items from commits. Having one item that repeats what other items on the same level say feels odd.

LaakkonenJussi:
> 
> The easiest to trace are the commits with "(cherry picked from commit ...)" note as the `-x` option to `git cherry-pick` does :)
> 

This approach has been never used with this fork. At this point it might not be any additional benefit until we update to some baseline release from upstream.

> 
> Yeah, summary in PR message is great, but keep in mind that if you write it as a changelog item, then it will appear in changelog on the same level as the items from commits. Having one item that repeats what other items on the same level say feels odd.

I'm well aware of this. It really does not bother me.

PR #36, opened 2022-12-19
[vpn] Connecting fixes, fix VPNC connect reply and improve logging. Fixes JB#55105

LaakkonenJussi:
Add logging of unhandled errors with VPNC to help in error resolving.

Fix the replying to the connect request when VPNC connection abruptly fails
because of server/client misconfiguration. This avoids the D-Bus LimitsExceeded
error when VPNC keeps trying and failing by always replying to the connection
request.

Use natt as default NAT Traversal mode. This is the default according to man
pages.

Use EALREADY in provider.c connect reply for VPNs that were already connecting.
This helps in differentiate between the two cases: VPN was just connected or
has been trying to connect and waiting for vpnd reply. Also connect_timeout()
is set only once per connection attempt.


pvuorela (approved):
Seems simple.

Thaodan (approved):

pvuorela:
Got quite a bit more stuff since my earlier approve. Not that familiar with the details and don't have vpnc to test with, but can't say there's anything wrong either so guess this is still approved from my side.

connman/src/service.c
```diff
@@ -8389,10 +8390,12 @@ int __connman_service_connect(struct connman_service *service,
 
 	service->connect_reason = reason;
 
-	if (err >= 0)
+	switch (err) {
+	case 0:
```
llewelld, line 8394: Previously this function returned 0 here if `err > 0`, but now it will continue instead. I couldn't figure out whether you can assume that `err <= 0`, or if this change is intentional, so I'm flagging it just in case.

llewelld (commented):
The changes look sensible, just one point that I wasn't sure about.

I'm also not able to test this unfortunately. Are there any publicly accessible VPNC servers I could try with? Otherwise, it looks sensible enough, I would be comfortable to approve it anyway.

connman/src/service.c
LaakkonenJussi: In any case success should be a return of 0. So this previously did not react to any other negative error than `-EINPROGRESS` and the resulting calls to [service.c:service_connect()](https://github.com/sailfishos/connman/blob/master/connman/src/service.c#L8229)  leading to two: 

1. [network.c:__connman_network_connect()](https://github.com/sailfishos/connman/blob/master/connman/src/network.c#L1533) -> calls to plugin that do not seem to use positive ERRNO values for any other than D-Bus messages, check with:
  ```
  for i in $(errno --list|cut -d' ' -f1) ; do
      find connman/plugins/* -type f -exec grep -e $i {} + |grep -v -e "-$i" -e "DEFAULT"
  done
  ```
or

2. the call to [provider.c:__connman_provider_connect()](https://github.com/sailfishos/connman/blob/master/connman/src/provider.c#L313) -> [plugins/vpn.c:connect_provider()/provider_connect()](https://github.com/sailfishos/connman/blob/master/connman/plugins/vpn.c#L603)

Do return only negative values with error cases and 0 is a-ok. In any ways treating any other than 0 as a mark of success is wrong in this context. I cannot remember a case where that was used as success but there have been issues of handling the non-negative ERRNOs.

This could for sure handle more errors but maybe for later..

llewelld: Thanks for checking all this so thoroughly. I'm reassured that it's okay.

llewelld (approved):
I wasn't able to test the changes I'm afraid, but the code looks good to me (and builds fine, apart from the unrelated failing unit test raised elsewhere).

PR #38, opened 2023-01-05
Synchronize state enums and fix state transitioning

LaakkonenJussi:
Revisit of JB#59447 after comments to RFC:

Synchronize the state enums with service state enums.

Remove support to reverse state back to ASSOCIATION.

Add full state machine support for VPN agentless VPNs.


LaakkonenJussi:
Commits with only `[]` in title will be included to later formed upstream patch set.

connman/vpn/vpn-provider.h
```diff
@@ -179,6 +179,7 @@ struct vpn_provider_driver {
 	int (*route_env_parse) (struct vpn_provider *provider, const char *key,
 			int *family, unsigned long *idx,
 			enum vpn_provider_route_type *type);
+	int (*uses_vpn_agent) (struct vpn_provider *provider);
```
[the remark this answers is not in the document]
LaakkonenJussi: Yeah, true. Will make a small change and force push.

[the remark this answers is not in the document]
LaakkonenJussi: Done. Resolving.

pvuorela (approved):
Don't know the code around here too well, but suppose the other current reviewers don't know it better either so just approving now. At least it appears sane.

connman/vpn/vpn-provider.c
```diff
@@ -1713,8 +1713,19 @@ int __vpn_provider_connect(struct vpn_provider *provider, DBusMessage *msg)
...
+	if (err == -EINPROGRESS) {
+		vpn_provider_set_state(provider,
+					VPN_PROVIDER_STATE_ASSOCIATION);
+
+		/*
+		 * If the VPN does not use VPN agent do direct transition to
+		 * connect in order to support the complete state machine.
+		 */
+		if (provider->driver && provider->driver->uses_vpn_agent &&
+				provider->driver->uses_vpn_agent(provider))
```
llewelld, line 1725: Is there anywhere to record the semantics of `uses_vpn_agent`? I had a quick look but didn't see anywhere that looked appropriate, but maybe I missed it.

If I understand correctly, if I'm writing a vpn driver and define a `uses_vpn_agent` function, then a zero return value means "use VPN agent" and a non-zero return value means "skip VPN agent".

And presumably the idea is that this matches with the general connman semantics of "zero means success"?

llewelld (approved):
The code all looks good to me. I didn't have anything very valuable to say, but managed to make two comments anyway ;)

connman/vpn/vpn-provider.c
LaakkonenJussi: Zero or non-existing function means that VPN agent is used. But then again, using `bool` here would be self-explanatory. I'll change to that.

connman/vpn/plugins/vpn.c
```diff
@@ -766,6 +757,27 @@ static int vpn_route_env_parse(struct vpn_provider *provider, const char *key,
...
+
+	name = vpn_provider_get_driver_name(provider);
+	vpn_driver_data = g_hash_table_lookup(driver_hash, name);
+
+	if (vpn_driver_data && vpn_driver_data->vpn_driver->uses_vpn_agent)
+		return vpn_driver_data->vpn_driver->uses_vpn_agent(provider);
+
+	/*
+	 * The use of VPN agent must be explicitly dropped. Default to that
+	 * VPN agent is in use when the function is not implemented.
```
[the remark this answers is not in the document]
LaakkonenJussi: I guess your version improves this. Will change.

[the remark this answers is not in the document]
llewelld: Thanks for the change. There's certainly not much in it either way, and I feel this is a tricky thing to express succinctly.

connman/vpn/vpn-provider.c
llewelld: I'm probably misunderstanding, but I would expect `uses_vpn_agent()` returning `true` to mean "stay on `VPN_PROVIDER_STATE_ASSOCIATION` and don't go directly on to `VPN_PROVIDER_STATE_CONNECT`".

If that's the intention, I think the function should be negated:
```suggestion
				!provider->driver->uses_vpn_agent(provider))
```
Sorry to dwell on this, I just want to make sure I'm understanding.

LaakkonenJussi: I think I had a stroke or something or it is https://github.com/sailfishos/connman/pull/39 ... Yeah, there should be a negation there

LaakkonenJussi: Changed with 1567662d64749df840efb4319ba357d435740a5d.

llewelld:
Thanks for the most recent changes; still looks good to me.

PR #40, opened 2023-01-17
[connman] Apply regdom changes to powering up devices. Fixes JB#55556

LaakkonenJussi:
The devices, let alone technologies are not loaded nor set up yet when timezone
reads the current time and attempts to set up regulatory domain. This change
fixes the issue by saving the regulatory domain as a global for the
technologies which is then used by the device during power up.

### DIFF
--- a/connman/src/connman.h
+++ b/connman/src/connman.h
@@ -672,6 +672,7 @@ void __connman_technology_remove_interface(enum connman_service_type type,
 				int index, const char *ident);
 void __connman_technology_notify_regdom_by_device(struct connman_device *device,
 						int result, const char *alpha2);
+const char *__connman_technology_get_regdom(enum connman_service_type type);
 const char *__connman_technology_get_tethering_ident(
 					struct connman_technology *tech);
 enum connman_service_type __connman_technology_get_type(
--- a/connman/src/device.c
+++ b/connman/src/device.c
@@ -589,6 +589,7 @@ int connman_device_set_powered(struct connman_device *device,
 						bool powered)
 {
 	enum connman_service_type type;
+	const char *alpha2;
 
 	DBG("device %p powered %d", device, powered);
 
@@ -606,6 +607,15 @@ int connman_device_set_powered(struct connman_device *device,
 	if (!device->powered) {
 		__connman_technology_disabled(type);
 		return 0;
+	} else {
+		/*
+		 * Check if technology has regdom set and apply it This may
+		 * have been changed when the device was powered off and, thus
+		 * the new regdom has not been set.
+		 */
+		alpha2 = __connman_technology_get_regdom(type);
+		if (alpha2)
+			connman_device_set_regdom(device, alpha2);
 	}
 
 	__connman_technology_enabled(type);
--- a/connman/src/technology.c
+++ b/connman/src/technology.c
@@ -50,6 +50,8 @@ static unsigned int global_offlinemode_override; /* Technology bitmask */
 struct connman_access_tech_policy *tech_access_policy;
 static unsigned int enable_delayed_ids[MAX_CONNMAN_SERVICE_TYPES] = { 0 };
 
+static char *global_regdom = NULL;
+
 struct connman_rfkill {
 	unsigned int index;
 	enum connman_service_type type;
@@ -366,7 +368,13 @@ int connman_technology_set_regdom(const char *alpha2)
 					driver->set_regdom(technology, alpha2);
 			}
 		}
+
+		/* Save regdom for this technology */
+		connman_technology_regdom_notify(technology, alpha2);
 	}
+
+	g_free(global_regdom);
+	global_regdom = g_strdup(alpha2);
 
 	return 0;
 }
@@ -385,6 +393,22 @@ static struct connman_technology *technology_find(enum connman_service_type type
 	}
 
 	return NULL;
+}
+
+const char *__connman_technology_get_regdom(enum connman_service_type type)
+{
+	struct connman_technology *technology;
+
+	DBG("type %d/%s", type, get_name(type));
+
+	technology = technology_find(type);
+	if (!technology)
+		return NULL;
+
+	if (technology->regdom)
+		return technology->regdom;
+
+	return global_regdom;
 }
 
 bool connman_technology_get_wifi_tethering(const char **ssid,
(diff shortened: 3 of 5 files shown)


pvuorela:
Installed and booted, but I still seem to have US returned on iw reg get.

LaakkonenJussi:
> Installed and booted, but I still seem to have US returned on iw reg get.

Did you have WLAN powered? It was also in the old code that the change is done only for powered devices.

On Xperias there usually is 2 interfaces shown, the topmost is the global which is affected by this setting.

And logs from, e.g., adding `CONNMAN_ARGS=-d src/timezone.c -d src/technology.c -d src/device.c` to `/etc/sysconfig/connman` will tell more.

LaakkonenJussi:
After boot with this change on X10 III and WLAN powered I do get:
```
[root@Xperia10III ~]# iw reg get
global
country FI: DFS-ETSI
	(2402 - 2482 @ 40), (N/A, 20), (N/A)
	(5170 - 5250 @ 80), (N/A, 23), (N/A), AUTO-BW
	(5250 - 5330 @ 80), (N/A, 23), (0 ms), DFS, AUTO-BW
	(5490 - 5710 @ 160), (N/A, 30), (0 ms), DFS
	(5850 - 5870 @ 10), (N/A, 30), (N/A)
	(5860 - 5880 @ 10), (N/A, 30), (N/A)
	(5870 - 5890 @ 10), (N/A, 30), (N/A)
	(5880 - 5900 @ 10), (N/A, 30), (N/A)
	(5890 - 5910 @ 10), (N/A, 30), (N/A)
	(5900 - 5920 @ 10), (N/A, 30), (N/A)
	(5910 - 5930 @ 10), (N/A, 30), (N/A)
	(57000 - 66000 @ 2160), (N/A, 40), (N/A)
```

pvuorela:
WLAN is on and used.

Adding some logs I'm getting at least this
```
Jan 19 13:33:56 Xperia10II-DualSIM connmand[28536]: src/timezone.c:__connman_timezone_lookup() sysconfig zone UTC
Jan 19 13:33:56 Xperia10II-DualSIM connmand[28536]: src/timezone.c:__connman_timezone_lookup() localtime zone Europe/Mariehamn
Jan 19 13:33:56 Xperia10II-DualSIM connmand[28536]: src/timezone.c:get_timezone_alpha2() read /usr/share/zoneinfo/zone1970.tab for Europe/Mariehamn
```

Timezone is Europe/Helsinki both in UI and target for symbolic link in /var/lib/timed/localtime

LaakkonenJussi:
> WLAN is on and used.
> 
> Adding some logs I'm getting at least this
> 
> ```
> Jan 19 13:33:56 Xperia10II-DualSIM connmand[28536]: src/timezone.c:__connman_timezone_lookup() sysconfig zone UTC
> Jan 19 13:33:56 Xperia10II-DualSIM connmand[28536]: src/timezone.c:__connman_timezone_lookup() localtime zone Europe/Mariehamn
> Jan 19 13:33:56 Xperia10II-DualSIM connmand[28536]: src/timezone.c:get_timezone_alpha2() read /usr/share/zoneinfo/zone1970.tab for Europe/Mariehamn
> ```
> 
> Timezone is Europe/Helsinki both in UI and target for symbolic link in /var/lib/timed/localtime

Well, there is no entry for Europe/Mariehamn in the file: `grep -e Mariehamn /usr/share/zoneinfo/zone1970.tab` .

ConnMan compares the link set in
```
ls -la /var/lib/timed/localtime
lrwxrwxrwx    1 defaultu defaultu        35 Feb 14 10:11 /var/lib/timed/localtime -> /usr/share/zoneinfo/Europe/Helsinki
```
So if that points to `/usr/share/zoneinfo/Europe/Mariehamn` for reasons unknown I have no idea what to do in such case.

This PR does apply the proper regdom when everything is correct. However, dealing with cases such as:
```
grep -e Helsinki /usr/share/zoneinfo/zone1970.tab
FI,AX	+6010+02458	Europe/Helsinki
```

Needs further work. I'd put this in and do a continuation task for those odd cases.

rainemak:
> Needs further work. I'd put this in and do a continuation task for those odd cases.

Sounds like a good approach.

LaakkonenJussi:
> 
> Timezone is Europe/Helsinki both in UI and target for symbolic link in /var/lib/timed/localtime

Actually there is a bug in timed. If you manually pick Helsinki and change to Mariehamn the symlink is not changed. Same the other way around.

To overcome this we chould use the `/usr/share/timezone/zone.tab` instead of `zone1970.tab` because the former one has each of the timezones listed with a single ISO3661 code. It seems to work, at least on X10 III the WLAN adapter does recognize `AX` as Finnish regdom. I'll make the minimal change as the both of those `*.tab` come from [the same source](https://www.iana.org/time-zones).

pvuorela:
> Actually there is a bug in timed. If you manually pick Helsinki and change to Mariehamn the symlink is not changed. Same the other way around.

So per what I tested earlier I wasn't ever having the link pointing to Mariehamn. Using zone.tab feels a bit suspicious since it's marked to be a deprecated version.

LaakkonenJussi:
> > Actually there is a bug in timed. If you manually pick Helsinki and change to Mariehamn the symlink is not changed. Same the other way around.
> 
> So per what I tested earlier I wasn't ever having the link pointing to Mariehamn. Using zone.tab feels a bit suspicious since it's marked to be a deprecated version.

And provide proof it has been deprecated.

pvuorela:
> No idea what you've been doing with your device but that is what ConnMan uses to determine the set TZ. Period. In your case it points to Mariehamn quite clearly. No need to argue about that.

At the moment I can only refer to the earlier "Timezone is Europe/Helsinki both in UI and target for symbolic link in /var/lib/timed/localtime". It's still pointing to Helsinki.

Perhaps I can debug this some day now further.

> And provide proof it has been deprecated.

```
$ head -1 /usr/share/zoneinfo/zone.tab 
# tzdb timezone descriptions (deprecated version)
```

LaakkonenJussi:
From zone1970.tab:
> # If a timezone covers multiple countries, the most-populous city is used,
> # and that country is listed first in column 1; any other countries
> # are listed alphabetically by country code.

So it is omitting a lot of data. In this case that file does not work as there is no other place to map the <continent>/<city> to a ISO3166 code used for WLAN regulatory domain.

zone.tab also says:
> # From Paul Eggert (2018-06-27):
> # This file is intended as a backward-compatibility aid for older programs.
> # New programs should use zone1970.tab.

rainemak (approved):
Given that this works now better. Let's have a go with it. Approving.

We can change back to the other zone file when it gets removed or so.

PR #45, opened 2023-06-08
[connman] Fix 64bit arch build warnings and deprecated func. Fixes JB#54603

LaakkonenJussi:
Fix the warnings 64bit arch build gives. Also replace deprecated g_memdup() with the new g_memdup2().

### DIFF
--- a/connman/unit/test-globalproxy.c
+++ b/connman/unit/test-globalproxy.c
@@ -289,7 +289,7 @@ DAPolicy* da_policy_new_full(const char* spec, const DA_ACTION* actions)
 // Needed to control permissions independent of reality
 void da_policy_unref(DAPolicy* policy)
 {
-	g_assert_cmphex((uint)policy, ==, POLICY_FAKE_POINTER);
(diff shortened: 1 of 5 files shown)


connman/unit/test-globalproxy.c
```diff
@@ -289,15 +289,15 @@ DAPolicy* da_policy_new_full(const char* spec, const DA_ACTION* actions)
 // Needed to control permissions independent of reality
 void da_policy_unref(DAPolicy* policy)
 {
-	g_assert_cmphex((uint)policy, ==, POLICY_FAKE_POINTER);
```
spiiroin, line 292: nit: While 'size_t' might silence warnings in this kinds of situations in some environments... there are 'intptr_t' and 'uintptr_t' c types that are meant for cases like this. Or, maybe even better, since this is glib code: GPOINTER_TO_UINT() macro could be used?

spiiroin (approved):

connman/unit/test-globalproxy.c
LaakkonenJussi: Actually `GPOINTER_TO_UINT()`  is ideal here. Changed and force-pushed.

PR #49, opened 2023-07-05
[connman] Cleanup post-install script. Fixes JB#60966

rainemak:
We have done multiple stop releases since this cleanup script was introduced. Thus, this can be safely removed.

See 3.0.2 release / sha1 cea1f1fd21b8b70dd70


LaakkonenJussi (approved):
LGTM

PR #52, opened 2023-12-21
Restructuring and fixes after the CLAT dual-index support changes

LaakkonenJussi:
These are some fixes made after testing and looking at the CLAT changes that enabled dual-index support. 

A kind of WIP still as waiting for community reports and fixes to those would be included here as well.

### DIFF
--- a/connman/src/connection.c
+++ b/connman/src/connection.c
@@ -443,9 +443,11 @@ static void set_default_gateway(struct gateway_data *data,
 
 	if (do_ipv4 && data->ipv4_gateway &&
 					data->ipv4_gateway->vpn) {
-		if (connman_inet_set_gateway_interface(data->index4) < 0)
-			DBG("cannot set VPN as gateway interface, index %d",
+		if (connman_inet_set_gateway_interface(data->index4) < 0) {
+			DBG("cannot set VPN v4 as gateway interface, index %d",
 							data->index4);
+			return;
+		}
 
 		data->ipv4_gateway->active = true;
 
@@ -454,16 +456,16 @@ static void set_default_gateway(struct gateway_data *data,
 			data->ipv4_gateway->vpn_phy_index,
 			data->ipv4_gateway->vpn_phy_ip);
 
-		__connman_service_indicate_default(data->service);
-
-		return;
+		goto done;
 	}
 
 	if (do_ipv6 && data->ipv6_gateway &&
 					data->ipv6_gateway->vpn) {
-		if (connman_inet_set_ipv6_gateway_interface(data->index6) < 0)
-			DBG("cannot set VPN as gateway interface, index %d",
+		if (connman_inet_set_ipv6_gateway_interface(data->index6) < 0) {
+			DBG("cannot set VPN v6 as gateway interface, index %d",
 							data->index6);
+			return;
+		}
 
 		data->ipv6_gateway->active = true;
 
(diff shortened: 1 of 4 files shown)


connman/src/network.c
```diff
@@ -215,15 +215,17 @@ static void dhcp_failure(struct connman_network *network,
...
 	if (ipconfig_ipv4 != ipconfig) {
 		DBG("ignore failure, ipconfig has changed");
 		return;
 	}
 
 	DBG("lease lost for ipconfig %p", ipconfig_ipv4);
 
-	if (!ipconfig_ipv4)
-		return;
+
```
mlehtima, line 228: Adds extra empty line.

LaakkonenJussi: Yep, missed that, thanks. Removed all extra empty lines and force-pushed.

connman/src/connection.c
```diff
@@ -480,19 +480,24 @@ static void set_default_gateway(struct gateway_data *data,
...
 							"0.0.0.0") == 0) {
-		if (connman_inet_set_gateway_interface(data->index4) < 0)
+		if (connman_inet_set_gateway_interface(data->index4) < 0) {
 			DBG("cannot set v4 gateway interface index %d",
 								data->index4);
+			return;
+		}
+
 		data->ipv4_gateway->active = true;
 		goto done;
```
spiiroin, line 490: out of curiousity: is there some deeper meaning between duplicated "indicate_default() + mid-function-return" code used above and this "goto where we have common indicate_default() + return" used elsewhere?

connman/src/service.c
```diff
@@ -9176,8 +9176,13 @@ void connman_service_create_ip4config(struct connman_service *service,
 {
 	DBG("ipv4 %p", service->ipconfig_ipv4);
 
-	if (service->ipconfig_ipv4)
+	if (service->ipconfig_ipv4) {
+		if (connman_ipconfig_get_index(service->ipconfig_ipv4) != index)
+			__connman_ipconfig_set_index(service->ipconfig_ipv4,
```
spiiroin, line 9181: Is there some benefit for condtitional set_index()? In other commit it was done unconditionally + set_index() is basically a logical nop if there is no index change.

connman/src/network.c
```diff
@@ -177,16 +177,16 @@ static void dhcp_success(struct connman_network *network,
 		goto err;
 
 	ipconfig_ipv4 = __connman_service_get_ip4config(service);
+	if (!ipconfig_ipv4)
+		return;
```
spiiroin, line 181: again, just wondering: here = ignored silently, below = debug logging -> intentional?

spiiroin (approved):

connman/src/network.c
LaakkonenJussi: This case is an error that can be silently ignored. AFAIK every service should have at least the ipconfig setup. The debug below is good for debugging cases where the ipconfig has been re-created before the DHCP returns. I've seen some rare cases happening like that and I left the debug for noticing the issue also via network.c.

connman/src/service.c
LaakkonenJussi: Well, maybe not. I'm just wondering if this would possibly break anything in some weird CLAT-scenario. However, the index should always be set to match the interface when the ipconfig is to be created. In many cases it may be set later on but that would be an explicit set and this just simply ensures that the index is always matched with the interface.

connman/src/connection.c
spiiroin: in any case: there is no behavioral difference, so this is not really relevant for review -> resolving

LaakkonenJussi: Did some additional restructuring after going through the code to make sure that the active is not anywhere needed in such cases. Made a consistency change to the code: 1abe9b6369c1f4d37223c4eb8b31249eb46c9874

Maybe it just removes some noise from the logs as the unset function is not triggered for an interface that is not actually set here.

connman/src/service.c
LaakkonenJussi: Removed that bit to skip the check. Yeah, it wasn't necessary. Overthinking again.

PR #55, opened 2024-01-25
Apply online check changes from upstream, add DNS routes for hostname lookup

LaakkonenJussi:
This applies many online check changes from upstream and adapts our fork into using the upstream approach. The approach is almost identical to ours and works in the same way.

Second thing is improving the online check robustness by making sure that it will be started always when a service enters ready state. There was an issue, most likely a race, when mobile data online check succeeds just before WLAN gets connected and into ready state the online check was not started for WLAN. The common, and upstream approach is to start the online check for the default service, which does not apply to us as on mobile devices there should be an online service as a default to pass traffic and only when the other, higher ranked, in this case WLAN service succeeds its online check it should become the new default.

Third addition is to add DNS routes for the DNS servers used by gweb when performing the hostname lookup for the online check url. Routes are added for the online check IP later and adding the DNS routes follows similar fashion. In order to be successful with this, and to online check to succeed as 404 is directly returned when hostname lookup fails, the old and stale gateway information must be cleared on the service along with other ipconfig cleanups. It can be left behind when the interface goes down before netlink can notify about route/gateway removal and when the interface goes down everything related is cleaned up by the kernel. Thus, no notify for ConnMan and adding a DNS routes fails with the previous WLAN's gateway.

[connman] Sync online check functionality with upstream. JB#61410

The commits from upstream:
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=396bb0c6d4bf40ca7f4061a175553fb866e8ff04
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=8688917d85ccab2ceed391315f2e56dd48036c35
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=e600366f603590b01449b34a8215efde54f566dd
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=69e9e25032946887aede3ebb6233739e7ceb2d49
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=c80aac568589154f7ff32081c2197e7fbdc0d4e8
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=69a221f912312d7a94eca40aa09a45e3bc70369a
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=a84f03f6f6c5d69cbb2e2c32326f9af603b0f8a7
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=4388028c4179362e2f409e6ab90d53975527cafe
https://git.kernel.org/pub/scm/network/connman/connman.git/commit/?id=abbb9eaafa99544c605dde92542d5561f639ed78

### DIFF
--- a/connman/gweb/gweb.c
+++ b/connman/gweb/gweb.c
@@ -99,6 +99,7 @@ struct web_session {
 	GWebResult result;
 
 	GWebResultFunc result_func;
+	GWebDnsRouteFunc dns_route_func;
 	GWebRouteFunc route_func;
 	GWebInputFunc input_func;
 	int fd;
@@ -107,6 +108,7 @@ struct web_session {
 	gpointer user_data;
 
 	bool cancelled;
+	bool dns_routes_set;
 };
 
 struct _GWeb {
@@ -473,6 +475,26 @@ static inline void call_result_func(struct web_session *session, guint16 status)
 
 	session->result_func(&session->result, session->user_data);
 
+}
+
+static inline int call_dns_route_func(struct web_session *session, bool add)
+{
+	if (add == session->dns_routes_set)
+		return -EALREADY;
+
+	if (session->dns_route_func) {
+		if (!session->dns_route_func(session->web->index, add,
+							session->user_data)) {
+			debug(session->web, "Failed to %s DNS routes",
+					add ? "add" : "remove");
+			// No service or gateway, route errors do not propagate
+			return -EINVAL;
+		}
+	}
+
+	session->dns_routes_set = add;
+
+	return 0;
 }
 
 static inline void call_route_func(struct web_session *session)
@@ -1276,6 +1298,11 @@ static void resolv_result(GResolvResultStatus status,
 					char **results, gpointer user_data)
 {
 	struct web_session *session = user_data;
+	int err;
+
+	err = call_dns_route_func(session, false);
+	if (err && err != -EALREADY)
+		debug(session->web, "failed to cleanup DNS routes");
 
 	if (session->cancelled)
 		return;
@@ -1310,7 +1337,8 @@ static bool is_ip_address(const char *host)
 static guint do_request(GWeb *web, const char *url,
 				const char *type, GWebInputFunc input,
 				int fd, gsize length, GWebResultFunc func,
-				GWebRouteFunc route, gpointer user_data)
+				GWebDnsRouteFunc dns_route, GWebRouteFunc route,
+				gpointer user_data)
 {
 	struct web_session *session;
 	const gchar *host;
@@ -1344,6 +1372,7 @@ static guint do_request(GWeb *web, const char *url,
 	session->web = web;
 
 	session->result_func = func;
+	session->dns_route_func = dns_route;
 	session->route_func = route;
 	session->input_func = input;
 	session->fd = fd;
@@ -1380,9 +1409,17 @@ static guint do_request(GWeb *web, const char *url,
 		session->address_action = g_timeout_add(0, already_resolved,
 							session);
 	} else {
+		int err = call_dns_route_func(session, true);
+		if (err && err != -EALREADY)
+			debug(session->web, "failed to add DNS routes");
+
 		session->resolv_action = g_resolv_lookup_hostname(web->resolv,
 					host, resolv_result, session);
 		if (session->resolv_action == 0) {
+			err = call_dns_route_func(session, false);
+			if (err && err != -EALREADY)
+				debug(session->web, "failed to del DNS routes");
+
 			free_session(session);
 			return 0;
 		}
@@ -1394,16 +1431,19 @@ static guint do_request(GWeb *web, const char *url,
 }
 
 guint g_web_request_get(GWeb *web, const char *url, GWebResultFunc func,
-		GWebRouteFunc route, gpointer user_data)
+		GWebDnsRouteFunc dnsroute, GWebRouteFunc route,
+		gpointer user_data)
 {
-	return do_request(web, url, NULL, NULL, -1, 0, func, route, user_data);
+	return do_request(web, url, NULL, NULL, -1, 0, func, dnsroute, route,
+			user_data);
 }
 
 guint g_web_request_post(GWeb *web, const char *url,
 				const char *type, GWebInputFunc input,
 				GWebResultFunc func, gpointer user_data)
 {
-	return do_request(web, url, type, input, -1, 0, func, NULL, user_data);
+	return do_request(web, url, type, input, -1, 0, func, NULL, NULL,
+			user_data);
 }
 
 guint g_web_request_post_file(GWeb *web, const char *url,
@@ -1421,7 +1461,7 @@ guint g_web_request_post_file(GWeb *web, const char *url,
 	if (fd < 0)
 		return 0;
 
-	ret = do_request(web, url, type, NULL, fd, st.st_size, func, NULL,
+	ret = do_request(web, url, type, NULL, fd, st.st_size, func, NULL, NULL,
 			user_data);
 	if (ret == 0)
 		close(fd);
@@ -1431,6 +1471,8 @@ guint g_web_request_post_file(GWeb *web, const char *url,
 
 bool g_web_cancel_request(GWeb *web, guint id)
 {
+	int err;
+
 	if (!web)
 		return false;
 
@@ -1440,6 +1482,11 @@ bool g_web_cancel_request(GWeb *web, guint id)
 
 	g_resolv_cancel_lookup(web->resolv, session->resolv_action);
 	session->cancelled = true;
+
+	/* Remove nameserver routes if added and not removed before cancel */
+	err = call_dns_route_func(session, false);
+	if (err && err != -EALREADY)
+		debug(session->web, "Could not remove nameserver routes");
 
 	return true;
 }
--- a/connman/src/main.c
+++ b/connman/src/main.c
@@ -54,6 +54,14 @@
 #define DEFAULT_UMASK (0077)
 #define DEFAULT_LOCALTIME "/etc/localtime"
 
+/*
+ * We set the integer to 1 sec so that we have a chance to get
+ * necessary IPv6 router advertisement messages that might have
+ * DNS data etc.
+ */
+#define DEFAULT_ONLINE_CHECK_INITIAL_INTERVAL 1
+#define DEFAULT_ONLINE_CHECK_MAX_INTERVAL 12
+
 #define MAINFILE "main.conf"
 #define CONFIGMAINFILE CONFIGDIR "/" MAINFILE
 
@@ -100,6 +108,8 @@ static struct {
 	bool enable_login_manager;
 	char *localtime;
 	bool regdom_follows_timezone;
+	unsigned int online_check_initial_interval;
+	unsigned int online_check_max_interval;
 } connman_settings  = {
 	.bg_scan = true,
 	.pref_timeservers = NULL,
@@ -123,6 +133,8 @@ static struct {
 	.enable_login_manager = false,
 	.localtime = NULL,
 	.regdom_follows_timezone = false,
+	.online_check_initial_interval = DEFAULT_ONLINE_CHECK_INITIAL_INTERVAL,
+	.online_check_max_interval = DEFAULT_ONLINE_CHECK_MAX_INTERVAL,
 };
 
 #define CONF_BG_SCAN                    "BackgroundScanning"
@@ -151,6 +163,9 @@ static struct {
 #define CONF_ENABLE_LOGIN_MANAGER       "EnableLoginManager"
 #define CONF_LOCALTIME                  "Localtime"
 #define CONF_REGDOM_FOLLOWS_TIMEZONE    "RegdomFollowsTimezone"
+
+#define CONF_ONLINE_CHECK_INITIAL_INTERVAL "OnlineCheckInitialInterval"
+#define CONF_ONLINE_CHECK_MAX_INTERVAL     "OnlineCheckMaxInterval"
 
 static const char *supported_options[] = {
 	CONF_BG_SCAN,
@@ -182,6 +197,8 @@ static const char *supported_options[] = {
 	CONF_ENABLE_LOGIN_MANAGER,
 	CONF_LOCALTIME,
 	CONF_REGDOM_FOLLOWS_TIMEZONE,
+	CONF_ONLINE_CHECK_INITIAL_INTERVAL,
+	CONF_ONLINE_CHECK_MAX_INTERVAL,
 	NULL
 };
 
@@ -367,7 +384,7 @@ static void parse_config(GKeyFile *config)
 	struct in_addr ip;
         char *vendor_class_id;
 	gsize len;
-	int timeout;
+	int integer;
 
 	if (!config) {
 		connman_settings.auto_connect =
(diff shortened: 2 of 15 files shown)


connman/src/service.c
```diff
@@ -2051,6 +2069,24 @@ static void print_service_list_debug()
 	}
 }
 
+<<<<<<< Updated upstream
```
LaakkonenJussi, line 2072: Apparently this missed my eyes. TODO for myself.

LaakkonenJussi: Done, was just waiting commit *rolleyes*.

connman/src/service.c
```diff
@@ -1917,13 +1873,70 @@ void __connman_service_nameserver_del_routes(struct connman_service *service,
...
+{
+	if (!service)
+		return;
+
+	if (service->online_timeout_ipv4) {
+		g_source_remove(service->online_timeout_ipv4);
+		service->online_timeout_ipv4 = 0;
+		connman_service_unref(service);
+	}
+	if (service->online_timeout_ipv6) {
```
spiiroin, line 1911: there is potential unref(service) above, is it still ok to deref service?

LaakkonenJussi: Our version `stop_recurring_online_check()` that was removed had the same approach, this is the upstream naming/style for the function.

LaakkonenJussi: The unref is done because in wispr when the online check is started for a IP family the service is referenced. So each IP family type needs to have this unref. The crashing issue was elsewhere in the upstream changes, and that was unnecessary unref, which was removed.

Resolving this.

connman/src/service.c
```diff
@@ -8076,22 +8085,33 @@ static void redo_wispr(struct connman_service *service, enum connman_ipconfig_ty
 		return;
 	}
 
+	DBG("Retrying %s WISPr for %p %s",
+		__connman_ipconfig_type2string(type),
+		service, service->name);
+
 	__connman_wispr_start(service, type);
+	connman_service_unref(service);
```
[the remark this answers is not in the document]
LaakkonenJussi: Removed.

connman/gweb/gweb.c
```diff
@@ -475,6 +477,30 @@ static inline void call_result_func(struct web_session *session, guint16 status)
 
 }
 
+static inline int call_dns_route_func(struct web_session *session, bool add)
+{
+	bool ret;
+
+	if (add == session->dns_routes_set)
+		return -EALREADY;
```
spiiroin, line 485: I'm wondering about return values from this function

Nothing to do, everything is already as requested -> EALREADY *error* -> sounds like every caller should be prepared to handle this special case - which most do not which then makes one wonder whether it is intentional or not.

connman/gweb/gweb.c
```diff
@@ -475,6 +477,30 @@ static inline void call_result_func(struct web_session *session, guint16 status)
...
+
+	if (session->dns_route_func) {
```
spiiroin, line 487: if route func is null, function returns "succeeded" without doing anything / even toggling the state -> a bit fishy.

connman/gweb/gweb.c
```diff
@@ -1380,9 +1413,18 @@ static guint do_request(GWeb *web, const char *url,
 		session->address_action = g_timeout_add(0, already_resolved,
 							session);
 	} else {
+		// TODO make to use return value
+		int err = call_dns_route_func(session, true);
+		if (err)
```
spiiroin, line 1418: why not: if (err && err != -EALREADY)

connman/gweb/gweb.c
```diff
@@ -1380,9 +1413,18 @@ static guint do_request(GWeb *web, const char *url,
...
+			debug(session->web, "failed to add DNS routes");
+
 		session->resolv_action = g_resolv_lookup_hostname(web->resolv,
 					host, resolv_result, session);
 		if (session->resolv_action == 0) {
+			err = call_dns_route_func(session, false);
+			if (err)
```
spiiroin, line 1425: if the earlier call_dns_route_func(true) had genuine failure, state is false -> this call_dns_route_func(false) will EALREADY fail -> diagnostic noise?

connman/src/main.c
```diff
@@ -815,6 +858,17 @@ bool connman_setting_get_bool(const char *key)
...
 }
 
+unsigned int connman_setting_get_uint(const char *key)
+{
+	if (g_str_equal(key, CONF_ONLINE_CHECK_INITIAL_INTERVAL))
+		return connman_settings.online_check_initial_interval;
+
+	if (g_str_equal(key, CONF_ONLINE_CHECK_MAX_INTERVAL))
+		return connman_settings.online_check_max_interval;
+
```
spiiroin, line 868: Now silently returns 0 for unknown keys. Would it make sense to have some diagnostic message here?

connman/src/main.c
```diff
@@ -815,6 +858,17 @@ bool connman_setting_get_bool(const char *key)
 	return false;
 }
 
+unsigned int connman_setting_get_uint(const char *key)
+{
+	if (g_str_equal(key, CONF_ONLINE_CHECK_INITIAL_INTERVAL))
```
spiiroin, line 863: g_str_equal() does not handle null args, do we know that "key" is always non-null?

connman/src/wispr.c
```diff
@@ -904,9 +936,35 @@ static bool wispr_portal_web_result(GWebResult *result, gpointer user_data)
...
+	if (!g_strcmp0(proxy, "DIRECT"))
+		return NULL;
+
+	if (!g_str_has_prefix(proxy, "PROXY "))
+		return NULL;
+
+	proxy_server = g_strdup(proxy + 6);
+
+	/* Use first proxy server */
+	for (char *c = proxy_server; *c != '\0'; ++c) {
```
spiiroin, line 952: there are standard functions like strchr(), strcspn(), strpbrk(), etc that could be used for finding 1st instance of ';' instead of a custom loop...

connman/src/wispr.c
```diff
@@ -904,9 +936,35 @@ static bool wispr_portal_web_result(GWebResult *result, gpointer user_data)
...
 }
 
+static char *parse_proxy(const char *proxy)
+{
+	char *proxy_server;
+
+	if (!g_strcmp0(proxy, "DIRECT"))
+		return NULL;
+
+	if (!g_str_has_prefix(proxy, "PROXY "))
```
spiiroin, line 946: Presumably proxy arg is always non-null?

connman/src/service.c
```diff
@@ -1896,18 +1879,66 @@ void __connman_service_nameserver_add_routes(struct connman_service *service,
...
+		if (type == CONNMAN_IPCONFIG_TYPE_ALL)
+			type = CONNMAN_IPCONFIG_TYPE_IPV6;
+		else
+			return;
+	}
+
+	if (refcount6 != -1 && refcount6 != 1) {
+		if (type == CONNMAN_IPCONFIG_TYPE_ALL) {
+			type = CONNMAN_IPCONFIG_TYPE_IPV4;
+		} else
```
spiiroin, line 1935: nit: braces for 'if' but not for 'else'

connman/src/service.c
```diff
@@ -8153,16 +8281,26 @@ int __connman_service_online_check_failed(struct connman_service *service,
...
 
-	if (*online_check_count > 1)
-		--(*online_check_count);
-	else
-		connman_warn("Online check failed for %p %s", service, service->name);
+	timeout = connman_wakeup_timer_add_seconds(*interval * *interval,
+				redo_func, connman_service_ref(service));
+	if (type == CONNMAN_IPCONFIG_TYPE_IPV4) {
+		if (service->online_timeout_ipv4)
+			g_source_remove(service->online_timeout_ipv4);
```
spiiroin, line 8288: does not clear service->online_timeout_ipv4?
ditto for ipv6 timeout below

spiiroin (approved):
FWIW could not spot any real issues.
Some remarks, mostly about potentially too much or too little diagnostics - I'll leave it up to you to check if there is something worth addressing there.

connman/src/main.c
LaakkonenJussi: I think there is a pattern for every `connman_setting_get_*` to just silently ignore unknown keys. None of those report anything. And besides, this comes from a upstream change directly. Lets not break the pattern. Each of the values anyways has a default value so I think the caller should then determine based on that, and make debug for invalid return or something.

connman/src/wispr.c
LaakkonenJussi: Didn't pay much attention to this upstream change. But yeah, that could be changed indeed.

connman/src/wispr.c
LaakkonenJussi: Part of upstream change. And yes, the calling function, [`proxy_callback()`](https://github.com/sailfishos/connman/pull/55/files#diff-b082b989c711bf47589cd0a7c33a91e9b550ba1ca13f17504efe6f13f22433b8R971) has check for `proxy` being NULL.

connman/src/service.c
LaakkonenJussi: Ah, forgot to clean up this one..

connman/src/service.c
LaakkonenJussi: Wouldn't that be kind of moot as it is being setup with the new timeout few lines after?

connman/gweb/gweb.c
LaakkonenJussi: Yeah, that would work :)

connman/gweb/gweb.c
LaakkonenJussi: Well, I find it useful to differentiate between ok return and nothing done return. This is a pattern in many places around ConnMan to return `EALREADY` for a no-op when states match. Adding handling to callers for this.

connman/gweb/gweb.c
LaakkonenJussi: Yeah, it should change state. Will change that one.

connman/gweb/gweb.c
LaakkonenJussi: Ah, right, need to handle that EALREADY here as well, thanks.

connman/src/service.c
LaakkonenJussi: change to [5f49f52](https://github.com/sailfishos/connman/pull/55/commits/5f49f5274da18de3a7193cfb058bcaf6e276fa05) cleans this up.

connman/gweb/gweb.c
LaakkonenJussi: This [a4e3c22](https://github.com/sailfishos/connman/pull/55/commits/a4e3c2202bab7ec34a107eac78bab9c12b6567ff) should handle the issue here.

connman/gweb/gweb.c
LaakkonenJussi: Done in [a4e3c22](https://github.com/sailfishos/connman/pull/55/commits/a4e3c2202bab7ec34a107eac78bab9c12b6567ff)

connman/gweb/gweb.c
LaakkonenJussi: Changed with [a4e3c22](https://github.com/sailfishos/connman/pull/55/commits/a4e3c2202bab7ec34a107eac78bab9c12b6567ff)

connman/gweb/gweb.c
LaakkonenJussi: Handled in [a4e3c22](https://github.com/sailfishos/connman/pull/55/commits/a4e3c2202bab7ec34a107eac78bab9c12b6567ff)

connman/src/service.c
spiiroin: hmm, yes, my pattern matching misfired -> is ok as-is

connman/src/main.c
LaakkonenJussi: I think this is good since the developer would instantly see that `NULL` keys are causing instant crashes. Kind of ensures that code remains correct, as searching for uses of `connman_settings_get_` shows. Lets leave as is.

connman/src/wispr.c
LaakkonenJussi: Like this 36e30b88e6f01782ae8a669fed4bb62212d8bafa

connman/src/service.c
```diff
@@ -5018,9 +5029,28 @@ static DBusMessage *set_property(DBusConnection *conn,
...
+			/*
+			 * After the user set ns routes have been added restore
+			 * old refcounters to avoid situation where online
+			 * check might remove the routes if this change happens
+			 * during the initial resolving of the online check URL.
+			 * Set the old value only for the IP type for which the
+			 * routes were added.
+			 */
+			if (ipv4_refcount_old && gw_family == AF_INET)
+				service->nameservers_ipv4_refcount =
```
spiiroin, line 5061: If I understand what is going on...
- at start: cache original refcounts
- earlier: ensure that both refcounts=1 or 0 + del_routes() -> both refcounts==0
- now, right after add_routes(): either ipv4 or ipv6 refcount == 1 (and the other is zero)
- we want that the refcount that is non-zero to be restored to original value

Now, I wonder why one would use test that feels like it is based on assumptions on what ought to have happened instead of just checking what happened? I mean, could this be simply:

    if (service->nameservers_ipvN_refcount > 0)
        service->nameservers_ipvN_refcount = ipvN_refcount_old;

connman/src/service.c
```diff
@@ -8430,6 +8449,8 @@ int __connman_service_ipconfig_indicate_state(struct connman_service *service,
 	case CONNMAN_SERVICE_STATE_IDLE:
 	case CONNMAN_SERVICE_STATE_FAILURE:
 		__connman_ipconfig_disable(ipconfig);
+		service->nameservers_ipv4_refcount = 0;
+		service->nameservers_ipv6_refcount = 0;
```
spiiroin, line 8456: just wondering: having seen that __connman_service_nameserver_del_routes() turns into nop on zero refcounts... is it ok to just clear these without further actions/checks?

connman/src/service.c
LaakkonenJussi: This is because the type of the nameserver routes added are determined by the `gw` family, here the `gw_family` type. Well, thinking that way cleans up the code and ends up reliable result. Good suggestion, I'll change that.

connman/src/service.c
LaakkonenJussi: I was looking for a place to ensure the reset of these values since at this point the service is anyways completely down and should not have any routes set anymore so it would be safe to clear these just to be sure that they start from 0 when connecting the service. I'll need to check if this happens at any tests with some added debug.

connman/src/service.c
LaakkonenJussi: Changed with b38de1d56f3095940f4a92f0eb5709bd086609de, which is going to be squeezed into some other commit later.

connman/gweb/gweb.c
```diff
@@ -475,6 +477,30 @@ static inline void call_result_func(struct web_session *session, guint16 status)
...
+
+		if (!ret) {
+			debug(session->web, "Failed to %s DNS routes",
+					add ? "add" : "remove");
+			// no service or gateway, route errors do not propagate
+			return -EINVAL;
+		}
+	}
+
+	session->dns_routes_set = ret;
```
LaakkonenJussi, line 499: Hmh, whoops, this `ret` does not have a value if there is no route function set.

LaakkonenJussi: Better to drop that `bool ret` and use the `add` value as it is now, to set the requested value.

PR #57, opened 2024-02-23
Upstream fixes: prefixlen for routes, support subnet routes, fix inet_pton retval tests

LaakkonenJussi:
This syncs inet.c a bit closer to upstream by applying:
 - Prefixlen for adding a route
 - Support for subnet routes in
 - Fix return value tests for inet_pton

### DIFF
--- a/connman/src/inet.c
+++ b/connman/src/inet.c
@@ -3424,7 +3424,7 @@ static int ipv6_neigbour_proxy(int index, bool enable, const char *ipv6_address,
 	attr->rta_type = NDA_DST;
 	attr->rta_len = RTA_LENGTH(sizeof(struct in6_addr));
 
-	if (!inet_pton(AF_INET6, ipv6_address, RTA_DATA(attr))) {
+	if (inet_pton(AF_INET6, ipv6_address, RTA_DATA(attr)) != 1) {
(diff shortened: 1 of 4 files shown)


mlehtima (approved):

connman/src/inet.c
```diff
@@ -599,6 +599,16 @@ int connman_inet_add_network_route_with_metric(int index, const char *host,
 	addr.sin_addr.s_addr = inet_addr(host);
 	memcpy(&rt.rt_dst, &addr, sizeof(rt.rt_dst));
 
+	/*
+	 * Don't add a routes for link-local or unspecified
+	 * destination address coupled with unspecified gateway.
+	 */
+	if ((!host || is_addr_ll(AF_INET, (struct sockaddr *)&addr) || __connman_inet_is_any_addr(host, AF_INET))
```
spiiroin, line 606: There is curious inconsistency between this function and ipv6 version. The latter null checks 'host' right at the start vs this one does not -> these added lines null check 'host' after it has already been used (ipv4) or when it has already been established that it is not null (ipv6).

spiiroin (approved):

connman/src/inet.c
LaakkonenJussi: Dropped that patch. Let it simmer in upstream still. Does not seem to be urgent for us now.

PR #59, opened 2024-02-27
Support --data-ciphers option in OpenVPN plugin

LaakkonenJussi:
This can be apparently used with --cipher option so no need to complicate with checks that which to use.


mlehtima (approved):

LaakkonenJussi:
Need to still add also support for the --data-ciphers-fallback option. It seems to be commonly used as well.

mlehtima (approved):

Thaodan (approved):
LGTM

PR #60, opened 2024-02-29
 Start service after quota to avoid fs read-only issues

LaakkonenJussi:
Set the drop-in which requires home to be mounted to order connmand (and connman-vpnd, which depends on connmand) to run after the quota service for home has been run. Quota service is triggered by home.mount and may make the fs as read-only for the period of quota check. This is an insurance to ensure that there will be no filesystem read-only issues regardless of what the quota service does.

### DIFF
--- a/rpm/01-require-home-mount-to-be-present.conf
+++ b/rpm/01-require-home-mount-to-be-present.conf
@@ -4,7 +4,14 @@
 # PartOf ensures that connman is stopped before home is unmounted. This is
 # because if the configurations go missing when connman is running undefined
 # behavior can happen.
+# Running the service after quota for home has been run ensures that regardless
+# of what the quota service has (as quotacheck tries to remount the filesystem
+# read-only before starting the scan) resulting in situations where fs cannot be
+# written to, which was especially case with vpnd running on slower, older
+# devices. Setting this for connmand will make connman-vpnd work as well since
+# it depends on connman.service to start prior to starting VPN service.
 
 [Unit]
 RequiresMountsFor=/home
 PartOf=home.mount
+After=quota@home.service


rpm/01-require-home-mount-to-be-present.conf
```diff
@@ -4,7 +4,14 @@
...
+# of what the quota service has (as quotacheck tries to remount the filesystem
+# read-only before starting the scan) resulting in situations where fs cannot be
+# written to, which was especially case with vpnd running on slower, older
+# devices. Setting this for connmand will make connman-vpnd work as well since
+# it depends on connman.service to start prior to starting VPN service.
 
 [Unit]
 RequiresMountsFor=/home
 PartOf=home.mount
+After=quota@home.service
```
keto, line 17: I think this might not be enough depending on how systemd interprets the quota service as started, i.e. the quotacheck might still be running... I'm not completely sure how that works.

And in general this is not very feasible solution as it would need to be applied to all services that expect home to be writable when the mount is ready. So instead the quota setup and (conditional) quotacheck should probably be tied somewhere earlier in the chain.

LaakkonenJussi: The real question is actually that why the read only filesystem error is reported via GLib also when the filesystem is writable. I think this works because the quota service has the `ExecStartPost` set, which is waited to finish before the startup of the service is complete:
`Note that the execution of ExecStartPost= is taken into account for the purpose of Before=/After= ordering constraints.` https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html

LaakkonenJussi: And that quotaon cannot be ran until the quotacheck finishes, right?

LaakkonenJussi: > And in general this is not very feasible solution as it would need to be applied to all services that expect home to be writable when the mount is ready. So instead the quota setup and (conditional) quotacheck should probably be tied somewhere earlier in the chain.

Not all, just a few. This is because many do depend on `network.target` that is ran after `connman.service` has been started. See the `systemd-analyze plot`:

image

LaakkonenJussi: And really to my surprize, the start times for both sailjaild and apkd were reduced, although they now start after quota, which always takes roughly the same time:
image

mlehtima: This could be a bit problematic on community devices which don't have a home mount and therefore no quota@home.service

LaakkonenJussi:
Closing since the issue is more general and is to be handled most likely with this as a best/simplest solution for now: https://github.com/sailfishos/quota/pull/2

PR #66, opened 2024-04-26
Connman upgrade to 1.33

LaakkonenJussi:


connman/gdbus/client.c
```diff
@@ -853,28 +853,30 @@ gboolean g_dbus_proxy_method_call(GDBusProxy *proxy, const char *method,
...
-		setup(&iter, data->user_data);
+		setup(&iter, user_data);
 	}
 
+	if (!function)
+		return g_dbus_send_message(client->dbus_conn, msg);
+
+	data = g_try_new0(struct method_call_data, 1);
+	if (data == NULL)
+		return FALSE;
```
spiiroin, line 873: after being moved, this mid-function return now leaks msg

connman/gdbus/client.c
```diff
@@ -853,28 +853,30 @@ gboolean g_dbus_proxy_method_call(GDBusProxy *proxy, const char *method,
...
+
+	data->function = function;
+	data->user_data = user_data;
+	data->destroy = destroy;
```
spiiroin, line 877: random observation: looks like  the idea is that caller yields ownership of user_data and expects that destroy callback will be called for it at some point. And there already were N ways for that not to happen (every mid-function return) => potential memory leaks. And now there is a new one: the freshly added shortcut for "we do not care about return message so lets skip user data cleanup too" case... what was the point of this commit again?

connman/gsupplicant/supplicant.c
```diff
@@ -2135,9 +2235,22 @@ static void interface_property(const char *key, DBusMessageIter *iter,
 	} else if (g_strcmp0(key, "Networks") == 0) {
 		supplicant_dbus_array_foreach(iter, interface_network_added,
 								interface);
-	} else
+	} else if (g_strcmp0(key, "DisconnectReason") == 0) {
+		int reason_code;
+		if (dbus_message_iter_get_arg_type(iter) != DBUS_TYPE_INVALID) {
```
spiiroin, line 2240: the usual: variable type is wrong and type test is interesting, but fingers crossed...

spiiroin (changes requested):

connman/gsupplicant/supplicant.c
LaakkonenJussi: Actually, we don't even use this internal g_supplicant.  We use gsupplicant from [libgsupplicant](https://github.com/sailfishos/libgsupplicant). We can ignore this, we are not going to switch back to this apparently and there are more fixes on this in the upstream when we upgrade further.

connman/gdbus/client.c
LaakkonenJussi: Anyways this kind of small allocations should take the whole ConnMan down according to [updated coding style](https://git.kernel.org/pub/scm/network/connman/connman.git/tree/doc/coding-style.txt#n158), so, well, perhaps we could add something here.

LaakkonenJussi: But if you really think about this. If a small memory allocation fails it reeks of a much bigger issue as memory is running out. Do you think the software will continue run after this? I think not because then there are gazillion other small reservations that start to fail and this just ends up dying at some other point. So memory leak in such case has no significance whatsoever. I personally do not see any use of putting any effort towards this.

The struct itself is not a really bug one, two function pointers and a void pointer. So according to the coding style this should be left to crash anyways.

connman/gdbus/client.c
LaakkonenJussi: I think this was made because someone in their plugin(s) did not want to pass a function to the `g_dbus_proxy_method_call()` since only thing using this is the Bluetooth plugin.

LaakkonenJussi: Ok, later on this is more extensively used by `iwd` plugin and systemd DNS resolver.

LaakkonenJussi: But was [3b0a203](https://github.com/sailfishos/connman/pull/66/commits/3b0a2036ecfa0647b90cf602ad19319a1c907d17) something you were after?

connman/gdbus/client.c
LaakkonenJussi: Well, it is at least in sync with the rest of the code after [3b0a203](https://github.com/sailfishos/connman/pull/66/commits/3b0a2036ecfa0647b90cf602ad19319a1c907d17)

connman/gdbus/client.c
spiiroin: It has been a while, so IIRC: yeah, the function had minor problems to start with, and now  after some changes it has new ones including possibility of becoming a memory leak generator even on a supposedly happy path.

connman/gdbus/client.c
```diff
@@ -869,17 +869,22 @@ gboolean g_dbus_proxy_method_call(GDBusProxy *proxy, const char *method,
...
 	data->user_data = user_data;
 	data->destroy = destroy;
 
-
 	if (g_dbus_send_message_with_reply(client->dbus_conn, msg,
 					&call, METHOD_CALL_TIMEOUT) == FALSE) {
 		dbus_message_unref(msg);
+
+		if (data->destroy)
+			data->destroy(data->user_data);
```
spiiroin, line 886: What I meant is that this sort of thing needs to be done unless data ownership is passed onward [=dbus_pending_call_set_notify() below] -> on *every* mid-function return.

spiiroin (approved):
All my concerns have been addressed

PR #72, opened 2024-06-26
Connman upgrade to 1.37

LaakkonenJussi:
Updated baseline to 1.37.

Has:
 - ACD (Address Collision Detection) support (not in use yet)
 - Tethering client list support for UI (not implemented to our WiFi yet)
 - Certain technologies can be set to autoconnect via .conf
 - NTP/timeserver improvements
 - DHCP related changes
 - Full scan enabling support for WiFi
 - Configuration option to treat gateways as timeservers (our `#define` use for this was removed)

And many other smaller ones.

### DIFF
--- a/connman/client/commands.c
+++ b/connman/client/commands.c
@@ -792,8 +792,6 @@ static void move_before_append_args(DBusMessageIter *iter, void *user_data)
 
 	dbus_message_iter_append_basic(iter,
 				DBUS_TYPE_OBJECT_PATH, &path);
-
-	return;
 }
 
 static int cmd_service_move_before(char *args[], int num,
@@ -852,8 +850,6 @@ static void move_after_append_args(DBusMessageIter *iter, void *user_data)
 
 	dbus_message_iter_append_basic(iter,
 				DBUS_TYPE_OBJECT_PATH, &path);
-
-	return;
 }
 
 static int cmd_service_move_after(char *args[], int num,
@@ -1930,8 +1926,6 @@ static void session_create_append_dict(DBusMessageIter *iter, void *user_data)
 
 		index += append.values;
 	}
-
-	return;
 }
 
 static void session_create_append(DBusMessageIter *iter, void *user_data)
--- a/connman/client/commands.c
+++ b/connman/client/commands.c
@@ -39,6 +39,7 @@
 #include "dbus_helpers.h"
 #include "input.h"
 #include "services.h"
+#include "tethering.h"
 #include "peers.h"
 #include "commands.h"
 #include "agent.h"
@@ -311,6 +312,18 @@ static int peers_list(DBusMessageIter *iter,
 {
 	if (!error) {
 		__connmanctl_peers_list(iter);
+		fprintf(stdout, "\n");
+	} else
+		fprintf(stderr, "Error: %s\n", error);
+
+	return 0;
+}
+
+static int tethering_clients_list(DBusMessageIter *iter,
+					const char *error, void *user_data)
+{
+	if (!error) {
+		__connmanctl_tethering_clients_list(iter);
 		fprintf(stdout, "\n");
 	} else
 		fprintf(stderr, "Error: %s\n", error);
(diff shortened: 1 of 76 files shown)


connman/client/commands.c
```diff
@@ -318,6 +319,18 @@ static int peers_list(DBusMessageIter *iter,
 	return 0;
 }
 
+static int tethering_clients_list(DBusMessageIter *iter,
+					const char *error, void *user_data)
```
rainemak, line 323: Unused variable user_data... Maybe same at upstream. Could cause issues with compiler.

connman/client/commands.c
```diff
@@ -639,6 +652,17 @@ static int cmd_tether(char *args[], int num, struct connman_option *options)
 	return tether_set(args[1], set_tethering);
 }
 
+static int cmd_tethering_clients(char *args[], int num, struct connman_option *options)
```
rainemak, line 655: args & options are not used. Maybe touching those variables.

connman/gsupplicant/supplicant.c
```diff
@@ -4635,7 +4704,9 @@ static void add_network_security_peap(DBusMessageIter *dict,
 
 	}
 
-	if (g_str_has_prefix(ssid->phase2_auth, "EAP-")) {
+	if(g_strcmp0(ssid->phase2_auth, "GTC") == 0 && g_strcmp0(ssid->eap, "ttls") == 0)
```
rainemak, line 4707:
```suggestion
	if (g_strcmp0(ssid->phase2_auth, "GTC") == 0 && g_strcmp0(ssid->eap, "ttls") == 0)
```

connman/include/log.h
```diff
@@ -60,8 +60,8 @@ struct connman_debug_desc {
 	const char *file;
 #define CONNMAN_DEBUG_FLAG_DEFAULT (0)
 #define CONNMAN_DEBUG_FLAG_PRINT   (1 << 0)
-//#define CONNMAN_DEBUG_FLAG_ALIAS   (1 << 1)
-//#define CONNMAN_DEBUG_FLAG_HIDE_NAME (1 << 2)
+#define CONNMAN_DEBUG_FLAG_ALIAS   (1 << 1)
+#define CONNMAN_DEBUG_FLAG_HIDE_NAME (1 << 2)
```
rainemak, line 64: Intentional change?

connman/plugins/wifi.c
```diff
@@ -1876,14 +1869,15 @@ static int wifi_scan(enum connman_service_type type,
 	if (wifi->tethering)
 		return -EBUSY;
 
-	if (type == CONNMAN_SERVICE_TYPE_P2P)
+	if (params->type == CONNMAN_SERVICE_TYPE_P2P)
```
rainemak, line 1872: Can params itself be null?

connman/include/log.h
LaakkonenJussi: Needed because of https://github.com/sailfishos/connman/pull/72/commits/a524e73ec70b72acbc7dabf2c5f2ea033d52ad8f - does not build without these. No idea why they were disabled in the first place.

connman/plugins/wifi.c
LaakkonenJussi: This is the upstream WiFi plugin we don't even build. But it cannot be `NULL` https://github.com/sailfishos/connman/pull/72/commits/e49a87134120b97b3f539e97d08fcdf71d2bb3ba

connman/client/commands.c
LaakkonenJussi: I think this is left there for future use. Need to check what is in the upstream now. But anyways, this is the `connmanctl` binary for accessing options, services etc. via cli.

LaakkonenJussi: Ah, no. This is what the function prototype has to have. It is a callback to D-Bus call. Should not be touched.

connman/client/commands.c
LaakkonenJussi: Same here, function prototype requires. Should not mess with it.

rainemak (approved):
Approving. I'd not mind if @mlehtima would give a look as well.

I'd still try to integration this rather sooner than later. If we found issues those could be fixed on top.

mlehtima (approved):

PR #76, opened 2024-09-05
Use GResolv for DNS reresolve in WireGuard VPN plugin

LaakkonenJussi:
In order to avoid blocking vpnd when network is not properly connected do the DNS reresolve with GResolve first, `getaddrinfo()` later.

### DIFF
--- a/connman/vpn/plugins/wireguard.c
+++ b/connman/vpn/plugins/wireguard.c
@@ -43,6 +43,8 @@
 #include <connman/setting.h>
 #include <connman/vpn-dbus.h>
 
+#include <gweb/gresolv.h>
+
 #include "../vpn-provider.h"
 #include "../vpn.h"
 
@@ -59,7 +61,10 @@ struct wireguard_info {
 	struct wg_peer peer;
 	char *endpoint_fqdn;
 	char *port;
-	int reresolve_id;
+	guint reresolve_id;
+	GResolv *resolv;
+	guint resolv_id;
+	guint remove_resolv_id;
 };
 
 struct sockaddr_u {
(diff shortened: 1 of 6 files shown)


connman/vpn/plugins/wireguard.c
```diff
@@ -338,6 +409,35 @@ static gboolean wg_dns_reresolve_cb(gpointer user_data)
...
+	if (!info->resolv) {
+		connman_error("cannot create GResolv");
+		return G_SOURCE_REMOVE;
+	}
+
+	info->resolv_id = vpn_util_resolve_hostname(info->resolv,
+						info->endpoint_fqdn,
+						resolve_endpoint_cb, info);
+
+	if (info->resolv_id <= 0) {
```
LaakkonenJussi, line 435: Hm, this is halfway useless check since it is a `guint`.

connman/vpn/plugins/wireguard.c
```diff
@@ -293,43 +298,109 @@ static bool sockaddr_cmp_addr(struct sockaddr_u *a, struct sockaddr_u *b)
...
+	case G_RESOLV_RESULT_STATUS_FORMAT_ERROR:
+	case G_RESOLV_RESULT_STATUS_NAME_ERROR:
+	case G_RESOLV_RESULT_STATUS_NOT_IMPLEMENTED:
+	case G_RESOLV_RESULT_STATUS_REFUSED:
+	case G_RESOLV_RESULT_STATUS_NO_ANSWER:
+		DBG("stop DNS reresolve");
+		if (info->provider)
+			vpn_provider_add_error(info->provider,
+					VPN_PROVIDER_ERROR_CONNECT_FAILED);
+
```
LaakkonenJussi, line 375: Not sure if this should just disconnect right away..

LaakkonenJussi:
Needs cleanup but should be now quite close to completion, so ready for reviews.

connman/vpn/plugins/wireguard.c
```diff
@@ -293,42 +298,89 @@ static bool sockaddr_cmp_addr(struct sockaddr_u *a, struct sockaddr_u *b)
...
+		DBG("resolv already removed");
+		return;
 	}
 
-	/* If this fails after being connected it means configuration error
-	 * that results in connection errors. */
-	err = parse_endpoint(info->endpoint_fqdn,
-			info->port, &addr);
+	info->remove_resolv_id = g_timeout_add(0, remove_resolv, info);
+	info->resolv_id = 0;
```
spiiroin, line 339: this looks suspicious. why is resolv_id just cleared here? surely if it is still active it should be canceled / cleared elsewhere / something?
or alternatively: makes me think whether that "resolv already removed" condition a bit earlier is correct or not?

connman/vpn/plugins/wireguard.c
```diff
@@ -293,42 +298,89 @@ static bool sockaddr_cmp_addr(struct sockaddr_u *a, struct sockaddr_u *b)
...
+
+static gboolean remove_resolv(gpointer user_data)
 {
 	struct wireguard_info *info = user_data;
-	struct sockaddr_u addr;
+
+	DBG("");
+
+	if (info->remove_resolv_id)
+		g_source_remove(info->remove_resolv_id);
```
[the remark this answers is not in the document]
spiiroin: ah. it is used both as callback and function - eww.

connman/vpn/plugins/wireguard.c
```diff
@@ -338,6 +409,35 @@ static gboolean wg_dns_reresolve_cb(gpointer user_data)
 			info->device.name);
 
 	run_dns_reresolve(info);
+}
+
+static gboolean wg_dns_reresolve_cb(gpointer user_data)
+{
```
spiiroin, line 415: NB info->reresolve_id associated with this callback is not cleared here, there and elsewhere...

connman/vpn/plugins/wireguard.c
```diff
@@ -346,6 +446,13 @@ static void run_dns_reresolve(struct wireguard_info *info)
 	if (info->reresolve_id)
 		g_source_remove(info->reresolve_id);
 
+	if (vpn_provider_get_connection_errors(info->provider) >=
+						DNS_RERESOLVE_ERROR_LIMIT) {
+		connman_warn("reresolve error limit reached");
+		vpn_died(NULL, -ENONET, info->provider);
+		return;
```
spiiroin, line 453: if reresolve_id was nonzero and thus removed, this early return leaves stale source id behind

connman/vpn/plugins/wireguard.c
spiiroin: also that special return from here just before doing the same return anyway seems a bit useless

connman/vpn/plugins/wireguard.c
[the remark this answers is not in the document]
LaakkonenJussi: Yes this is the same practice as in many places in ConnMan.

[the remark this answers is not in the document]
LaakkonenJussi: 2d6d484922e3631d6f9ce2b8d954bcdc42959dc3 does the split, resolving.

connman/vpn/vpn-util.c
```diff
@@ -221,3 +221,26 @@ int vpn_util_create_path(const char *path, uid_t uid, gid_t grp, int mode)
 	return err;
 }
 
+unsigned int vpn_util_resolve_hostname(GResolv *resolv,
+				const char *hostname, GResolvResultFunc func,
+				gpointer user_data)
+{
+	if (!resolv || !hostname || !func)
+		return 0;
```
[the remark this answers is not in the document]
LaakkonenJussi: Yeah, not needed I think. Let the gresolv.c to handle them: bc45c13c64cb3146d48e7aaaf83dc8f25ac66b5d

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Yes this needs clarification. The thing is that when the GResolv callback is returned, so the lookup is completed, GResolv cannot be unref'd within the callback. Therefore, the timeout function to call `remove_resolv()` is used, by zeroing the resolv_id so the lookup is not attempted to be canceled - as it completed. The resolv struct is manipulated by gresolv.c after the callback returns, it would segfault if done here.

A bit hacky, needs a better approach? Yes to both but I can live with this approach.

connman/vpn/plugins/wireguard.c
LaakkonenJussi: https://github.com/sailfishos/connman/blob/master/connman/gweb/gresolv.c#L1030 well, interesting, return is `guint` but it does return `-EIO` as well. Classic. I guess this needs to be addressed as well.

LaakkonenJussi: Oh well, a workaround to get the error and return 0 whenever an error happens: b857c8f6bdec18c090d3e15ab9e02038fdf674c4 - it now makes the VPN die as those errors are quite severe (EIO).

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Yeah, made this better with 9ce8543874c54a9b936850b0bdb1d9c1bf6d6cdb

connman/vpn/plugins/wireguard.c
LaakkonenJussi: 9ce8543874c54a9b936850b0bdb1d9c1bf6d6cdb I think this is now ok.

connman/vpn/plugins/wireguard.c
spiiroin: A bit hacky, yes. But with explanation understandable. I guess it is ok.

spiiroin (approved):

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Latest changes make it do so, the connect error limit takes care of it.

LaakkonenJussi:
Moving the other task commits away. This part was reviewed, merging.

PR #80, opened 2024-10-08
Make VPN autoconnect to disconnect the VPN, deny access to vpnd Connection/Disconnection

LaakkonenJussi:
VPNs did not get disconnected when autoconnect value was changed. This addresses the issue and is more in sync with the enabling of the autoconnect, which connects a VPN.

Prevent use of net.connman.Connection.Connect and Disconnect methods. It is better to use the main service API to do this, to avoid cases, like we now have, that Connect goes through connmand and Disconnect directly to vpnd.

### DIFF
--- a/connman/src/service.c
+++ b/connman/src/service.c
@@ -4900,6 +4900,30 @@ int connman_service_reset_ipconfig_to_address(struct connman_service *service,
 	return err;
 }
 
+// TODO: make this into a configuration option
+static bool autoconnect_controls_service(struct connman_service *service)
+{
+	if (!service)
+		return false;
+
+	switch (service->type) {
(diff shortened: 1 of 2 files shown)


LaakkonenJussi:
I was thinking should the access functionality to be extended to match a D-Bus sender name too in the permissions. But maybe that would be better done later in a task of its own.

LaakkonenJussi:
For example the `connect` command in the [`connmanctl` uses `net.connman` D-Bus service to connect](https://github.com/sailfishos/connman/blob/master/connman/client/commands.c#L737) and `net.connman.vpn` is used for parameters, listing etc. So the restriction based on the D-Bus name acquired through state check of connmand should be a secure way of verifying that which is the D-Bus name connmand uses.

spiiroin (approved):

PR #81, opened 2024-10-16
[vpn] Fix VPN PrefixLength use, handle CIDR in WireGuard. Fixes JB#62646

LaakkonenJussi:


connman/include/provider.h
```diff
@@ -74,6 +74,8 @@ struct connman_ipaddress;
 #define connman_provider_unref(provider) \
 	connman_provider_unref_debug(provider, __FILE__, __LINE__, __func__)
 
+#define AF_INET46 46
```
LaakkonenJussi, line 77: This is not the best way to do it, location should be elsewhere but this is only used in this scope. The `PF_MAX` is defined as 46...

connman/plugins/vpn.c
```diff
@@ -1591,6 +1590,31 @@ static void set_route(struct connection_data *data, struct vpn_route *route)
 static int save_route(GHashTable *routes, int family, const char *network,
 			const char *netmask, const char *gateway);
 
+
+static int set_network_route(struct connection_data *data, struct vpn_route *rt)
+{
+	int err;
+
+	if (!data|| !rt)
```
LaakkonenJussi, line 1598: Note to self: cleanup style

connman/plugins/vpn.c
```diff
@@ -1704,6 +1745,14 @@ static bool check_routes(struct connman_provider *provider)
 	return false;
 }
 
+static void set_default_route(struct connection_data *data, int family,
+							char *ipaddr_any)
+{
+	struct vpn_route def_route = {family, ipaddr_any, ipaddr_any,NULL};
```
LaakkonenJussi, line 1751: Note to self: cleanup style

connman/vpn/plugins/wireguard.c
```diff
@@ -226,50 +245,67 @@ static int parse_endpoint(const char *host, const char *port, struct sockaddr_u
...
-	tokens = g_strsplit(address, "/", -1);
-	if (g_strv_length(tokens) != 2) {
-		g_strfreev(tokens);
+	addresses = g_strsplit(address, ", ", -1);
+	if (!g_strv_length(addresses))
 		return -EINVAL;
-	}
 
-	prefixlen = g_ascii_strtoull(tokens[1], &end, 10);
+	for (i = 0; addresses[i]; i++) {
```
LaakkonenJussi, line 264: TODO: limit this to two addresses, IPv4 and IPv6. ConnMan has only one struct for each, so having more than 1 IPv4 and 1 IPv6 is not possible as of yet.

connman/include/provider.h
LaakkonenJussi: Changed to use a cleaner way. But then again, maybe VPNs should be modified to be similar to regular services. Not in this PR though, so resolving this and writing a task for that to be done later on.

connman/plugins/vpn.c
LaakkonenJussi: Moved to other PR.

mlehtima (approved):

PR #82, opened 2024-10-22
Split routing and route setup changes for WireGuard

LaakkonenJussi:
With WireGuard split routing option is to be overridden as the AllowedIPs define the IP ranges that can be used according to the cryptokey routing table set on the server. Therefore, the split routing option is set only when there are no "any" routes defined.

This also contains changes to add routes to the IP networks defined with AllowedIPs. For now, the routes are added only based on the IP family of the current endpoint. The routes are removed and re-added if the endpoint resolves to another address, and in such case it is done without a delay. Normally, at startup, there is a 200ms delay.

Also took one upstream commit on preventing to add invalid routes to routing table. Might be useful in some cases, maybe even here.

### DIFF
--- a/connman/vpn/plugins/wireguard.c
+++ b/connman/vpn/plugins/wireguard.c
@@ -53,6 +53,7 @@
 
 #define DNS_RERESOLVE_TIMEOUT 20
 #define DNS_RERESOLVE_ERROR_LIMIT 5
+#define ROUTE_SETUP_TIMEOUT 200 // ms
 #define ARRAY_SIZE(a) (sizeof(a)/sizeof(a[0]))
 
 struct wireguard_info {
@@ -66,6 +67,7 @@ struct wireguard_info {
 	guint resolv_id;
 	guint remove_resolv_id;
 	guint dying_id;
+	guint route_setup_id;
 };
 
 struct sockaddr_u {
@@ -262,6 +264,25 @@ struct wg_ipaddresses {
 	struct connman_ipaddress *ipaddress_ipv6;
 };
 
+static char *cidr_to_netmask(int family, unsigned char cidr)
+{
+	switch (family) {
+	case AF_INET:
+		return g_strdup_printf("%d.%d.%d.%d",
+				((0xffffffff << (32 - cidr)) >> 24) & 0xff,
+				((0xffffffff << (32 - cidr)) >> 16) & 0xff,
+				((0xffffffff << (32 - cidr)) >> 8) & 0xff,
+				((0xffffffff << (32 - cidr)) >> 0) & 0xff);
+
+	case AF_INET6:
+		return g_strdup_printf("%u", cidr);
+	default:
+		break;
+	}
+
+	return NULL;
+}
+
 static int parse_addresses(const char *address, const char *gateway,
 		struct wg_ipaddresses *ipaddresses)
 {
@@ -303,12 +324,7 @@ static int parse_addresses(const char *address, const char *gateway,
 			}
 
 			family = AF_INET;
-			netmask = g_strdup_printf("%d.%d.%d.%d",
-				((0xffffffff << (32 - prefixlen)) >> 24) & 0xff,
-				((0xffffffff << (32 - prefixlen)) >> 16) & 0xff,
-				((0xffffffff << (32 - prefixlen)) >> 8) & 0xff,
-				((0xffffffff << (32 - prefixlen)) >> 0) & 0xff);
-
+			netmask = cidr_to_netmask(family, prefixlen);
 			ipaddress = connman_ipaddress_alloc(family);
 			err = connman_ipaddress_set_ipv4(ipaddress, tokens[0],
 							netmask, gateway);
@@ -401,6 +417,7 @@ static bool sockaddr_cmp_addr(struct sockaddr_u *a, struct sockaddr_u *b)
 }
 
 static void run_dns_reresolve(struct wireguard_info *info);
+static void run_route_setup(struct wireguard_info *info, guint timeout);
 
 static void remove_resolv(struct wireguard_info *info)
 {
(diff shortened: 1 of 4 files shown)


connman/vpn/vpn-provider.c
```diff
@@ -3438,6 +3444,68 @@ int vpn_provider_append_route(struct vpn_provider *provider,
...
+
+int vpn_provider_delete_route(struct vpn_provider *provider,
+							unsigned long idx)
+{
+	struct vpn_route *route;
+
+	if (!__vpn_provider_check_routes(provider))
+		return -ENOENT;
+
+	route = g_hash_table_lookup(provider->routes, GINT_TO_POINTER(idx));
```
LaakkonenJussi, line 3497: Maybe drop this. Unnecessary.

connman/vpn/plugins/wireguard.c
```diff
@@ -501,6 +601,15 @@ static void run_dns_reresolve(struct wireguard_info *info)
 						wg_dns_reresolve_cb, info);
 }
 
+static void run_route_setup(struct wireguard_info *info, guint timeout)
+{
+	if (info->route_setup_id)
+		g_source_remove(info->route_setup_id);
+
+	info->route_setup_id = g_timeout_add_seconds(timeout, wg_route_setup_cb,
```
LaakkonenJussi, line 609: It might be better to have even shorter delay here for routes to be added, than one measured in seconds.

connman/vpn/vpn-provider.c
LaakkonenJussi: Dropped.

connman/vpn/plugins/wireguard.c
LaakkonenJussi: In ms now.

LaakkonenJussi:
Requries https://github.com/sailfishos/connman/pull/83 to be merged so the checks for adding routes based on configured IP families can be added here. Until that is done, this is in WIP status.

connman/vpn/plugins/wireguard.c
```diff
@@ -484,6 +515,70 @@ static gboolean wg_dns_reresolve_cb(gpointer user_data)
...
+	int family;
+	unsigned long idx = 0;
+
+	info->route_setup_id = 0;
+
+	// TODO use the peer.endpoint since it can be changed
+	family = connman_inet_check_ipaddress(info->endpoint_fqdn);
+	DBG("endpoint %s", info->endpoint_fqdn);
+
+	wg_for_each_allowedip(&info->peer, allowedip) {
```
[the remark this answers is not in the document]
LaakkonenJussi: Nope, it would really make a difference on an afterthought. The AllowedIPs dictate what is can be used.

connman/vpn/vpn-provider.c
```diff
@@ -3463,6 +3469,62 @@ int vpn_provider_append_route(struct vpn_provider *provider,
...
+
+	DBG("provider %p idx %lu family %d network %s netmask %s gateway %s",
+						provider, idx, family, network,
+						netmask, gateway);
+
+	route = provider_lookup_route(provider, idx, family);
+
+	route->network = g_strdup(network);
+	route->netmask = g_strdup(netmask);
+	route->gateway = g_strdup(gateway);
```
spiiroin, line 3487: Is there any chance that already set values would be overwritten here?

connman/vpn/plugins/wireguard.c
```diff
@@ -798,6 +923,9 @@ static int disconnect(struct vpn_provider *provider, int err)
 	if (info->reresolve_id > 0)
 		g_source_remove(info->reresolve_id);
 
+	if (info->route_setup_id > 0)
+		g_source_remove(info->route_setup_id);
```
spiiroin, line 927: Does it matter that stale ids are left in place?
Judging by what happens at run_dns_remove() before / after disconnect() call... bookkeeping looks fishy.

LaakkonenJussi: That case in `run_dns_remove()` is because disconnect happens from  within the plugin, and the `reresolve_id` is not going to be set. Also, in `disconnect()` the preparation for the plugin burial is started and when the`wg_died()` is called as a result, the struct holding these id's is free'd so it would be unnecessary to zero these out here.

connman/vpn/vpn-provider.c
LaakkonenJussi: This follows the same style that is used with other plugins, where they append the routes gradually. So, if the plugin does add the routes with the `idx` that is not going to be unique, which should never be done, it is doing this wrong and there will be leaks. The purpose here would be just add a new route, not sure if this needs to do it different from the other, to reply `-EALREADY` when a route with matching `idx`  is found.

LaakkonenJussi: Well, with 43ad8b65e480773bf208f535e85099ca4b71bac1 this is now a lot cleaner. Resolving this as.

connman/vpn/plugins/wireguard.c
spiiroin: to me, for this one timer id run_dns_reresolve(), looks like it does

```
        if (info->reresolve_id)
                g_source_remove(info->reresolve_id);
        if (info->reresolve_id > 0)
                g_source_remove(info->reresolve_id);
        info->reresolve_id = 0;

```
it is not going to cause crashes / like but... oh well, it is what it is

spiiroin (approved):

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Ok, I forgot to use Simo-English-Simo dictionary. Yeah, the order of the zeroing was faulty, fixed.

PR #83, opened 2024-10-22
Support dual IP addresses on VPNs

LaakkonenJussi:
This adds support for having both IPv4 and IPv6 on VPNs. Traditionally VPNs have been either or but WireGuard can utilize both.

Use a simple way of having the IP family types in an boolean array to avoid rewriting whole VPN implementation to match the other services. That work is to be done later.

### DIFF
--- a/connman/plugins/vpn.c
+++ b/connman/plugins/vpn.c
@@ -1590,55 +1590,92 @@ static void set_route(struct connection_data *data, struct vpn_route *route)
 static int save_route(GHashTable *routes, int family, const char *network,
 			const char *netmask, const char *gateway);
 
+
+static int set_network_route(struct connection_data *data, struct vpn_route *rt)
+{
+	int err;
+
+	if (!data || !rt)
+		return -EINVAL;
+
+	DBG("network %s gateway %s netmask %s for provider %p",
+					rt->network, rt->gateway, rt->netmask,
+					data->provider);
+
+	err = save_route(data->server_routes, rt->family, rt->network,
+						rt->netmask, rt->gateway);
+	if (err) {
+		connman_warn("failed to add network route for provider"
+					"%p", data->provider);
+		return err;
+	}
+
+	set_route(data, rt);
+
+	return 0;
+}
+
+static void free_network_route(struct vpn_route *rt)
+{
+	g_free(rt->network);
+	rt->network = NULL;
+
+	g_free(rt->netmask);
+	rt->network = NULL;
+
+	g_free(rt->gateway);
+	rt->network = NULL;
+}
+
 static int add_network_route(struct connection_data *data)
 {
 	struct vpn_route rt = { 0, };
-	int err;
+	int err = 0;
 
 	if (!data)
 		return -EINVAL;
 
-	rt.family = connman_provider_get_family(data->provider);
-	switch (rt.family) {
-	case PF_INET:
+	if (connman_provider_get_family(data->provider, AF_INET)) {
+		rt.family = AF_INET;
 		err = connman_inet_get_route_addresses(data->index,
-					&rt.network, &rt.netmask, &rt.gateway);
-		break;
-	case PF_INET6:
+					&rt.network, &rt.netmask,
+					&rt.gateway);
+		if (err) {
+			connman_error("cannot get IPv4 network/gateway/netmask "
+						"for %p/%s", data->provider,
+						data->ident);
+		} else {
+			err = set_network_route(data, &rt);
+			if (err)
+				connman_error("cannot set IPv4 network route "
+						"for %p/%s", data->provider,
+						data->ident);
+		}
+
+		free_network_route(&rt);
+	}
+
+	if (connman_provider_get_family(data->provider, AF_INET6)) {
+		rt.family = AF_INET6;
 		err = connman_inet_ipv6_get_route_addresses(data->index,
-					&rt.network, &rt.netmask, &rt.gateway);
-		break;
-	default:
-		connman_error("invalid protocol family %d", rt.family);
-		return -EINVAL;
+					&rt.network, &rt.netmask,
+					&rt.gateway);
+		if (err) {
+			connman_error("cannot get IPv6 network/gateway/netmask "
+						"for %p/%s", data->provider,
+						data->ident);
+		} else {
+			err = set_network_route(data, &rt);
+			if (err)
+				connman_error("cannot set IPv6 network route "
+						"for %p/%s", data->provider,
+						data->ident);
+		}
+
+		free_network_route(&rt);
 	}
 
-	DBG("network %s gateway %s netmask %s for provider %p",
-						rt.network, rt.gateway, rt.netmask,
-						data->provider);
-
-	if (err) {
-		connman_error("cannot get network/gateway/netmask for %p",
-							data->provider);
-		goto out;
-	}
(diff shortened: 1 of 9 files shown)


connman/src/shared/util.c
```diff
@@ -31,6 +31,8 @@
 
 #include "src/shared/util.h"
 
+#define ARRAY_SIZE(a) (sizeof(a)/sizeof(a[0]))
```
LaakkonenJussi, line 34: TODO: unused, remove. Cannot be used with pointers.

mlehtima (approved):

connman/vpn/vpn-provider.c
```diff
@@ -86,7 +87,7 @@ struct vpn_provider {
 	char *type;
 	char *host;
 	char *domain;
-	int family;
+	bool family[2];
```
spiiroin, line 90: As offsets to family array are defines (AF_INET6_POS etc) -> might be nice to have constant for array size too instead of hard coded 2?

Also: is initial state enforced somehow? Like by using calloc() / similar?

connman/plugins/vpn.c
```diff
@@ -1590,57 +1590,88 @@ static void set_route(struct connection_data *data, struct vpn_route *route)
...
+						data->ident);
+		else
+			err = set_network_route(data, &rt);
+
+		free_network_route(&rt);
+	}
+
+	if (connman_provider_get_family(data->provider, AF_INET6)) {
+		rt.family = AF_INET6;
+		err = connman_inet_ipv6_get_route_addresses(data->index,
```
spiiroin, line 1655: Success here can hide error from earlier op. Does it matter?

connman/src/connection.c
```diff
@@ -107,34 +107,40 @@ static struct gateway_data *lookup_gateway_data(struct gateway_config *config)
...
 	g_hash_table_iter_init(&iter, gateway_hash);
 
 	while (g_hash_table_iter_next(&iter, &key, &value)) {
 		struct gateway_data *data = value;
 
 		if (data->ipv4_gateway && data->index4 == index &&
 				g_str_equal(data->ipv4_gateway->gateway,
-					gateway))
+					gateway) &&
+				type == CONNMAN_IPCONFIG_TYPE_IPV4)
```
spiiroin, line 129: If 'type' is not ipv4 those other tests are meaningless -> you might want to do the type test first / something,

connman/unit/test-service.c
```diff
@@ -268,9 +268,9 @@ int connman_provider_get_index(struct connman_provider *provider)
 	return provider ? provider->index : -1;
 }
 
-int connman_provider_get_family(struct connman_provider *provider)
+bool connman_provider_get_family(struct connman_provider *provider, int family)
 {
-	return provider ? provider->family : PF_UNSPEC;
+	return provider && provider->family == family ? true : false;
```
spiiroin, line 273: nit: that "? true : false" part is superfuous

connman/vpn/plugins/wireguard.c
```diff
@@ -245,50 +245,93 @@ static int parse_endpoint(const char *host, const char *port, struct sockaddr_u
...
 	char **tokens;
 	char *end, *netmask;
 	int err;
+	int i;
 
-	tokens = g_strsplit(address, "/", -1);
-	if (g_strv_length(tokens) != 2) {
-		g_strfreev(tokens);
+	addresses = g_strsplit(address, ", ", -1);
+	if (!g_strv_length(addresses))
```
spiiroin, line 265: even array of size=zero takes memory -> mid-function return that leaks memory

connman/vpn/plugins/wireguard.c
```diff
@@ -245,50 +245,93 @@ static int parse_endpoint(const char *host, const char *port, struct sockaddr_u
...
+
+		DBG("address %s", addresses[i]);
+
+		prefixlen = g_ascii_strtoull(tokens[1], &end, 10);
+
+		if (inet_pton(AF_INET, tokens[0], buf) == 1) {
+			if (ipaddresses->ipaddress_ipv4) {
+				DBG("IPv4 address already set");
+				err = -EALREADY;
+				continue;
```
spiiroin, line 287: 'tokens' leaked

connman/vpn/plugins/wireguard.c
```diff
@@ -245,50 +245,93 @@ static int parse_endpoint(const char *host, const char *port, struct sockaddr_u
...
-	}
+			ipaddress = connman_ipaddress_alloc(family);
+			err = connman_ipaddress_set_ipv4(ipaddress, tokens[0],
+							netmask, gateway);
+			g_free(netmask);
+		} else if (inet_pton(AF_INET6, tokens[0], buf) == 1) {
+			if (ipaddresses->ipaddress_ipv6) {
+				DBG("IPv6 address already set");
+				err = -EALREADY;
+				continue;
```
spiiroin, line 305: 'tokens' leaked

spiiroin (changes requested):
Couple of possible memory leaks to check / fix.
Plus some style / eye of beholder things that do not really matter.

connman/vpn/vpn-provider.c
LaakkonenJussi: The structs are initialized with the zeroing calls, so yes. And the reset is done when the provider disconnects, on both.

But yeah, I'll replace the hardcoding, good point.

connman/plugins/vpn.c
LaakkonenJussi: Not really sure if it matters, if either of the routes gets set, the end result is half-correct if the VPN had both IP family routes to set.  Combining errors in these cases is not commonly used, there are some cases where either is ok = ok. Maybe just to report the error to logs instead of returning an error would be better approach if either fails?

connman/vpn/vpn-provider.c
LaakkonenJussi: 0bc55bf077789ce8551da1daddfd78ae90b693a5 remove hardcode.

connman/plugins/vpn.c
LaakkonenJussi: 08ede161a7003bcf4cce4b619ea3ed15be1df023 report better

connman/src/connection.c
LaakkonenJussi: Dropped.

connman/unit/test-service.c
LaakkonenJussi: 1ecd690ab4681cb5b7964339064d387b561eab38 cleanup

connman/vpn/plugins/wireguard.c
LaakkonenJussi: 417a130c62640aeb9daa5c1399d61039d4f04212 leak fixed

connman/vpn/plugins/wireguard.c
LaakkonenJussi: 417a130c62640aeb9daa5c1399d61039d4f04212 leak fixed

connman/vpn/plugins/wireguard.c
LaakkonenJussi: 417a130c62640aeb9daa5c1399d61039d4f04212 leak fixed

spiiroin (approved):

PR #88, opened 2025-01-10
Add conf.d support

LaakkonenJussi:


connman/src/firewall-iptables.c
```diff
@@ -2316,82 +2317,40 @@ static int init_dynamic_firewall_rules(const char *file)
...
-		DBG("no config dir %s", FIREWALLCONFIGDIR);
+	if (err) {
+		connman_warn("Failed to read main firewall config %s",
+							FIREWALLCONFIGFILE);
+		return err;
 	}
 
-	/* Error loading main configuration */
-	if (err)
+	err = util_read_config_files_from(FIREWALLCONFIGDIR, FIREWALLFILE,
```
direc85, line 2329: The config file suffix used here is `FIREWALLFILE` which expands to `firewall.conf`. Same in line 2677 too. Should these be `CONFIGSUFFIX` instead for `.conf` suffix?

connman/src/firewall-iptables.c
```diff
@@ -2697,98 +2657,72 @@ static int enable_new_firewall_rules(struct connman_service *service,
...
+	case -ENOMEM:
+		connman_warn("Failed to read configs from %s: %s",
+						FIREWALLCONFIGDIR,
+						strerror(-err));
+		return err;
+	case -ENOTDIR:
+	case -EACCES:
+		connman_warn("%s: %s, continue reloading firewall",
+						FIREWALLCONFIGDIR,
+						strerror(-err));
```
spiiroin, line 2690: after seeing N of these: Is there some profound reason why it would be preferable to have this kind of code duplication everywhere instead of doing such logging from within util_read_config_files_from()?

connman/src/shared/util.c
```diff
@@ -167,3 +171,113 @@ void util_reset_afs(bool *afs)
...
+	if (!files)
+		return 0;
+
+	files = g_list_sort(files, (GCompareFunc)g_strcmp0);
+
+	if (cb) {
+		for (iter = files; iter; iter = iter->next) {
+			filepath = iter->data;
+
+			err = cb(filepath);
```
spiiroin, line 270: Just wondering, not knowing what happens in existing config reader functionality: if a value is set in multiple files, it gets now "executed" several times. Any chance that such thing could trigger adverse things?

spiiroin (approved):

LaakkonenJussi:
I have some bigger fish to fry with this still, the main.c config is just so difficult to work with this kind of change I rewrote all of it and need to check if it works, so, maybe this was too early out of WIP.

connman/src/firewall-iptables.c
LaakkonenJussi: It was an old idea to avoid them getting mixed with normal configs, but they are in different directories now. So it was intentional choice back then and would need config renaming now. Maybe we keep it? It was documented I think.

connman/src/firewall-iptables.c
LaakkonenJussi: Well yeah the warnings can be moved, however this is the reload case that is different to normal loading here. Maybe that upper warning needs a better description. Not sure what to change, the different cases here kind of justify this..

connman/src/main.c
```diff
@@ -358,397 +88,52 @@ static void check_config(GKeyFile *config)
...
-			}
-		}
-		if (!found && !supported_options[i])
-			connman_warn("Unknown option %s in %s",
-						keys[j], MAINFILE);
+		bool found = __connman_setting_is_supported_option(keys[j]);
+		if (found)
+			break;
+
+		if (!found)
```
LaakkonenJussi, line 103: Todo: remove this `if`.

connman/src/main.c
```diff
@@ -1173,7 +395,8 @@ int main(int argc, char *argv[])
 	__connman_provider_init();
 	__connman_network_init();
 	__connman_config_init();
-	__connman_device_init(option_device, option_nodevice);
+	__connman_device_init(connman_setting_get_string("device"),
```
LaakkonenJussi, line 398: Todo: use the CONF_OPTION... values here and in the next

connman/src/setting.c
```diff
@@ -0,0 +1,1098 @@
...
+}
+
+typedef gboolean (*str_check_callback) (const char *value);
+typedef char** (*str_list_parse_callback) (char **str_list, gsize *len);
+typedef void (*str_list_callback) (const char *value);
+
+static void set_str(char **ptr, char *value, str_check_callback cb)
+{
+	if (*ptr)
+		g_free(*ptr);
```
LaakkonenJussi, line 588: Todo: set to NULL

connman/src/setting.c
```diff
@@ -0,0 +1,1098 @@
...
+						group, key, &len, &error);
+		if (!error) {
+			if (list_cb) {
+				set_str_list_cb(str_list, list_cb, len, append);
+				g_strfreev(str_list);
+			} else if (parse_cb) {
+				char **new_str_list = parse_cb(str_list, &len);
+				g_strfreev(str_list);
+
+				if (new_str_list && len)
```
LaakkonenJussi, line 859: Maybe the `NULL` check is enough here...

connman/src/shared/util.c
```diff
@@ -167,3 +171,113 @@ void util_reset_afs(bool *afs)
...
+
+	if (!g_file_test(path, G_FILE_TEST_IS_DIR))
+		return -ENOTDIR;
+
+	dir = g_dir_open(path, 0, &error);
+	if (!dir) {
+		if (error) {
+			connman_warn("cannot open dir %s, error: %s", path,
+								error->message);
+			err = -error->code;
```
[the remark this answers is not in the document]
LaakkonenJussi: Yeah these were completely different errors, also part of old code that can be fixed now.

connman/src/firewall-iptables.c
direc85: I looked at the config files on some device of mine and noticed that it's intentional, I was thinking it was a typo or something. Let's keep it like this, although the suffix is getting redundant as a side effect.

connman/src/shared/util.c
LaakkonenJussi: I think this is up to the processing function how to handle. There are no `+=` indicators for basic ini-style file processing so the value defined in the last config read overrides anything defined before. Each processing function must handle the cases of overriding, and later appending on some, like on lists. It is hard to make foolproof out of this, not sure if dropping of the file from the list in case of certain error should be done...

connman/src/shared/util.c
```diff
@@ -167,3 +171,113 @@ void util_reset_afs(bool *afs)
...
+		 * are added to avoid unnecessary reads of already read
+		 * configurations. Sort list after all are added.
+		 */
+		files = g_list_prepend(files, filepath);
+	}
+
+	g_dir_close(dir);
+
+	if (!files)
+		return 0;
```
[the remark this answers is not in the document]
LaakkonenJussi: Removed this. It will set what is in `files` to the `conffiles` if requested. Yeah, this is a kind of a case that could be rethought, like with the firewall-iptables.c that how to handle reloading. Worthy of an own task because this is already too big.

connman/src/shared/util.c
```diff
@@ -167,3 +171,113 @@ void util_reset_afs(bool *afs)
 
 	afs[AF_INET_POS] = afs[AF_INET6_POS] = false;
 }
+
+static bool is_file_symlink(const char *filename)
+{
+	int fd;
+
+	fd = open(filename, O_WRONLY | O_NOFOLLOW | O_CLOEXEC);
+	if (fd == -1 && errno == ELOOP)
```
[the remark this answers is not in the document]
LaakkonenJussi: Actually, this is where I apparently got the idea from in the first place long time ago: https://docs.gtk.org/glib/func.file_test.html 
```
// DON'T DO THIS
 if (!g_file_test (filename, G_FILE_TEST_IS_SYMLINK))
   {
     fd = g_open (filename, O_WRONLY);
     // write to fd
   }

 // DO THIS INSTEAD
 fd = g_open (filename, O_WRONLY | O_NOFOLLOW | O_CLOEXEC);
 if (fd == -1)
   {
     // check error
     if (errno == ELOOP)
       // file is a symlink and can be ignored
     else
       // handle errors as before
   }
 else
   {
     // write to fd
   }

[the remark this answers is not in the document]
LaakkonenJussi: Not sure why changing this to `g_file_test(filename, G_FILE_TEST_IS_SYMLINK)` breaks iptables unit tests and because it is quite laborious and boring to work with that I'd rather not go back there...

connman/src/shared/util.c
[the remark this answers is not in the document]
LaakkonenJussi: Added a GFileError converter.

direc85 (approved):
I've been test driving this, and poked it around and as far as I can tell it works well. Great work, huge effort!

PR #96, opened 2025-02-18
Fix some Clang analyzer findings

LaakkonenJussi:
These fixes are for some Clang analyzer findings. There are still more, some false positives and relying on memory allocation failing or does not understand the unref-approach used in many places, but these are most obvious ones for now.


direc85 (approved):
Good to have these fixes in!

A quick smoke test also passed on a device (data transfer over LTE and WLAN, developer mode).

mlehtima (approved):

PR #101, opened 2025-04-03
Fix VPN DNS leaks with systemd-resolved

LaakkonenJussi:
This fixes the leaking of DNS requests outside VPN with systemd-resolved. Only the online service's DNS are used, and the default one is set as the default DNS route link in resolved too, and disconnected are reverted to original state so they are not used. But most importantly, VPNs do not leak the queries outside.

### DIFF
--- a/connman/src/dns-systemd-resolved.c
+++ b/connman/src/dns-systemd-resolved.c
@@ -36,7 +36,7 @@
 #define SYSTEMD_RESOLVED_SERVICE "org.freedesktop.resolve1"
 #define SYSTEMD_RESOLVED_PATH "/org/freedesktop/resolve1"
 
-struct mdns_data {
+struct resolved_dbus_data {
 	int index;
 	bool enabled;
 };
@@ -48,6 +48,7 @@ static GDBusProxy *resolved_proxy;
 
 /* update after a full set of instructions has been received */
 static guint update_interfaces_source;
+static gint current_default_index = -1;
 
 struct dns_interface {
 	GList *domains;
@@ -55,6 +56,7 @@ struct dns_interface {
 	int index;
 	bool needs_domain_update;
 	bool needs_server_update;
+	bool enabled;
 };
 
 static gboolean compare_index(gconstpointer a, gconstpointer b)
@@ -405,7 +407,7 @@ static int setup_resolved(void)
 }
 
 static void setlinkmulticastdns_append(DBusMessageIter *iter, void *user_data) {
-	struct mdns_data *data = user_data;
+	struct resolved_dbus_data *data = user_data;
 	char *val = "no";
 
 	if (data->enabled)
@@ -419,7 +421,7 @@ static void setlinkmulticastdns_append(DBusMessageIter *iter, void *user_data) {
 
 int __connman_dnsproxy_set_mdns(int index, bool enabled)
 {
-	struct mdns_data data = { .index = index, .enabled = enabled };
+	struct resolved_dbus_data data = { .index = index, .enabled = enabled };
 
 	if (!resolved_proxy)
 		return -ENOENT;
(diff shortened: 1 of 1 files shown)


connman/src/dns-systemd-resolved.c
```diff
@@ -434,6 +445,288 @@ int __connman_dnsproxy_set_mdns(int index, bool enabled)
...
+
+	iface->enabled = enable;
+
+	if (!enable) {
+		DBG("disabling interface %d", iface->index);
+		resolved_disable_interface(iface->index);
+	} else {
+		DBG("re-enabling interface %d", iface->index);
+		set_systemd_resolved_values(iface);
+		if (is_empty(iface)) {
```
LaakkonenJussi, line 563: This cannot happen, because of the earlier check. I think this just needs to be done earlier..

connman/src/dns-systemd-resolved.c
```diff
@@ -434,6 +445,288 @@ int __connman_dnsproxy_set_mdns(int index, bool enabled)
...
+	int err;
+
+	DBG("enabled %d", enabled);
+
+	search_data.enable = enabled;
+	search_data.forced = true;
+
+	g_hash_table_foreach(interface_hash, check_interface, &search_data);
+
+	err = resolved_set_default_route(current_default_index, false);
```
LaakkonenJussi, line 601: Maybe this should be disabled only when offline mode is enabled.

connman/src/dns-systemd-resolved.c
```diff
@@ -434,6 +445,288 @@ int __connman_dnsproxy_set_mdns(int index, bool enabled)
...
+
+	if (current_default_index != index) {
+		err = resolved_set_default_route(current_default_index, false);
+		if (err && err != -ENODEV)
+			DBG("failed to unset %d as default",
+							current_default_index);
+
+		err = resolved_set_default_route(index, true);
+		if (err && err != -ENODEV)
+			DBG("failed to set %d as default", index);
```
LaakkonenJussi, line 670: Maybe the `current_default_index` should be cleared here. Or before.

connman/src/dns-systemd-resolved.c
```diff
@@ -434,6 +436,336 @@ int __connman_dnsproxy_set_mdns(int index, bool enabled)
...
+			iface->domains = NULL;
+
+			g_list_free_full(iface->servers, g_free);
+			iface->servers = NULL;
+		}
+
+		toggle_interface_dns(iface, search_data->enable);
+	/* If DNSs of an interface are to be enabled disable the others */
+	} else if (search_data->enable) {
+		DBG("disabling interface %d", iface->index);
```
mlehtima, line 587: Maybe useless logging, toggle_interface_dns always logs similar information.

connman/src/dns-systemd-resolved.c
```diff
@@ -434,6 +436,336 @@ int __connman_dnsproxy_set_mdns(int index, bool enabled)
...
+
+	return false;
+}
+
+static void toggle_interface_dns(struct dns_interface *iface, bool enable)
+{
+	DBG("%s interface %p/%d", enable ? "enable" : "disable", iface,
+						iface ? iface->index : -1);
+
+	if (!iface)
```
mlehtima, line 539: Code calling this function assumes already that iface is non-null.

connman/src/dns-systemd-resolved.c
LaakkonenJussi: True.

connman/src/dns-systemd-resolved.c
LaakkonenJussi: Yeah, I think I used this from other places before and just forgot this. The interface in hash table should not ever be `NULL` here.

LaakkonenJussi: 83e9a5782d65049693e9c11aa049623c853fdea3

mlehtima (approved):

PR #102, opened 2025-04-10
Add option for setting default mDNS value for all new Ethernet/WLAN services

LaakkonenJussi:
Add option to settings to define a default value for mDNS and document it.

Since the "false" mDNS value is dropped from the service file, so there is no knowledge if the user has disabled it explicitly and that is why the value is set only for new services. Otherwise each load of the service would set the system override for old services, from which mDNS was disabled - assuming system setting is "true".

Adding only to Ethernet and WLAN for now.

This has changes from the other PRs https://github.com/sailfishos/connman/pull/100 and https://github.com/sailfishos/connman/pull/101 included to ease testing.

### DIFF
--- a/connman/include/setting.h
+++ b/connman/include/setting.h
@@ -67,6 +67,7 @@ extern "C" {
 #define CONF_ENABLE_LOGIN_MANAGER            "EnableLoginManager"
 #define CONF_LOCALTIME                       "Localtime"
 #define CONF_REGDOM_FOLLOWS_TIMEZONE         "RegdomFollowsTimezone"
+#define CONF_DEFAULT_MDNS_CONFIGURATION      "DefaultmDNSConfiguration"
 
 #define CONF_ONLINE_CHECK_INITIAL_INTERVAL   "OnlineCheckInitialInterval"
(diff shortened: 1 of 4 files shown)


connman/include/setting.h
```diff
@@ -67,6 +67,7 @@ extern "C" {
 #define CONF_ENABLE_LOGIN_MANAGER            "EnableLoginManager"
 #define CONF_LOCALTIME                       "Localtime"
 #define CONF_REGDOM_FOLLOWS_TIMEZONE         "RegdomFollowsTimezone"
+#define CONF_DEFAULT_MDNS_CONFIGURATION      "DefaultmDNSConfiguration"
```
LaakkonenJussi, line 70: Not sure if we'd need two options:
 1. A system wide setting overriding _all_ Ethernet/WLAN service mDNS
 2. Default setting for the new services, which is much more lighter option. 
 
However, these might then conflict, global being off but new getting enabled, or vice versa. But that use case might be possible, like with ours, that do not touch old ones but keep the setting on new ones. We can live with this current approach for now and add more functionality later.

mlehtima (approved):

connman/include/setting.h
LaakkonenJussi: We live with this for now.

PR #107, opened 2025-06-09
Jb63544

LaakkonenJussi:


connman/src/iptables.c
```diff
@@ -2282,28 +2285,45 @@ static int iptables_replace(struct connman_iptables *table,
...
+	if (err) {
+		DBG("Getting iptables lock failed.");
+		if (err == -ENOLCK)
+			DBG("Timeout not implemented - fail miserably");
+
+		return -EINVAL;
+	}
+
+	err = setsockopt(table->ipt_sock, level, optname,
+				(level == IPPROTO_IP ? r->r : r->r6), optlen);
```
LaakkonenJussi, line 2323: TODO do this differently to avoid warning at build.

connman/src/shared/util.c
```diff
@@ -344,3 +345,67 @@ int util_read_config_files_from(const char *path, const char *suffix,
 
 	return 0;
 }
+
+int util_lock_file(int *lock_fd, const char *lock_path)
+{
+	if (!lock_path)
+		return -EINVAL;
+
+	/* TODO: check for file existense too -> fail lock automatically? */
```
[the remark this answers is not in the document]
LaakkonenJussi: Yeah, true. Removed the note.

connman/src/iptables.c
LaakkonenJussi: Either casting into `void *ptr` or doing it later as `(void*)r->r` ... Not sure which is the cleaner, but this avoids having to call lock separately.

LaakkonenJussi: a748a42d6e2a311be26589a251adf90374c59125 might be cleaner.

PR #108, opened 2025-09-09
Systemd resolved changes

LaakkonenJussi:


connman/src/service.c
```diff
@@ -1553,8 +1556,26 @@ static int nameserver_add_all(struct connman_service *service,
...
+	} else if (!fallback_dns) {
+		int index = __connman_service_get_index(service);
+		const char *dns = __connman_ipconfig_get_gateway_from_index(
+								index, type);
+
+		DBG("No DNS servers or fallback DNS servers set."
+						"Use gw %s as DNS", dns);
+		__connman_service_nameserver_append(service, dns, false);
+
+		nameserver_add(service, type, dns);
```
mlehtima, line 1573: Are both of these really needed? Based on quick look __connman_service_nameserver_append does call nameserver_add but with CONNMAN_IPCONFIG_TYPE_ALL not type, so is the forcing of type needed also?

LaakkonenJussi: It seems to be that latter can be dropped. The `type` is just to avoid adding the nameserver for a different IP family, or to avoid a check with early bail-out in `nameserver_available()`. Anyways, the IP family of the nameserver is checked there as well, so `nameserver_add()` is not needed here.

connman/src/shared/dns.c
```diff
@@ -2466,8 +2475,12 @@ static void destroy_listener(struct dns_listener_data *ifdata)
 	int index = connman_inet_ifindex("lo");
 
 	if (ifdata->index == index && !ifdata->lo_exclude) {
-		__connman_resolvfile_remove(index, NULL, "127.0.0.1");
-		__connman_resolvfile_remove(index, NULL, "::1");
+#ifdef SYSTEMD_RESOLVED_DNS_BACKEND
+		__connman_resolvfile_remove(index, NULL, DNS_BACKEND_V4);
+#else
+		__connman_resolvfile_remove(index, NULL, DNS_BACKEND_V4);
```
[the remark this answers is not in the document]
LaakkonenJussi: Yeah, it can be like that. Changed with #ifndef SYSTEMD_RESOLVED_DNS_BACKEND

PR #123, opened 2026-03-16
Add sae_pwe and sae_check_mfp support to settings and WiFi plugin

LaakkonenJussi:
Requires https://github.com/sailfishos/libgsupplicant/pull/6

### DIFF
--- a/connman/src/config.c
+++ b/connman/src/config.c
@@ -81,6 +81,8 @@ struct connman_config_service {
 	char **search_domains;
 	char **timeservers;
 	char *domain_name;
+	char *wpa3_sae_pwe;
+	bool wpa3_sae_check_mfp;
 };
 
 struct connman_config {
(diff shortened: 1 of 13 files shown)


direc85 (approved):

connman/src/config.c
```diff
@@ -282,6 +288,7 @@ static void unregister_service(gpointer data)
 	g_free(config_service->config_entry);
 	g_free(config_service->virtual_file);
 	g_free(config_service);
```
abranson: Shouldn't this one be freed last?

connman/src/service.c
```diff
@@ -1232,6 +1241,16 @@ static int service_save(struct connman_service *service)
 			service->private_key_passphrase);
 		set_config_string(keyfile, service->identifier,
 			PROP_PHASE2, service->phase2);
+		set_config_string(keyfile, service->identifier,
+			PROP_WPA3_SAE_PWE, service->wpa3_sae_pwe);
+		/* Save only when set, defaults to false */
+		if (service->wpa3_sae_check_mfp)
```
abranson, line 1246: I think there might be some confusion between unset and explicit false here. False will unset. Is that what should happen?

connman/src/config.c
LaakkonenJussi: Ah, yeah, true. Fixed, thanks.

connman/src/service.c
LaakkonenJussi: I guess the comment might be just misleading here. I'll remove it. That is the default behavior on boolean options in settings anyways. False is the default in practically everything as well as with this one, so it should not exists then.

abranson: Ok, I was worried it might want to combine with another setting elsewhere and get lost.

PR #124, opened 2026-03-25
Use _psk suffix for all WPA-supporting WLAN networks to avoid duplicate named networks

LaakkonenJussi:
Use the `_psk` for all WiFi networks having any WPA-security. This avoids having duplicate networks in the list if the networks are extended with other APs. To achieve this the BSSID creation, update or change does also the security parameter update. The real, accurate, WPA-type is used for D-Bus and for the WiFi driver.


direc85 (approved):
Tested on device and seems to work as expected.

Nice to have the RSSI adjusted a bit as well; haven't really paid attention to how my phones act, but now when I tested it it switched to a stronger network nice and early before the signal got too bad.

PR #125, opened 2026-03-30
Allocate the struct option for iptables 1.8.11 and above

LaakkonenJussi:
Iptables 1.8.11 now always frees the struct option passed via `xtables_merge_options()`. Allocate the struct in that case.


direc85 (approved):
Tested on device (with the iptables PR) and connections, tethering etc. work fine.

PR #127, opened 2026-04-13
Persist offline power policy per tech.

abranson:
If a technology is powered on in offline mode, remember that for next time. Initial behaviour is to turn off for everything except Ethernet and GPS, which aren't radio transmitting technologies.

### DIFF
--- a/connman/doc/technology-api.txt
+++ b/connman/doc/technology-api.txt
@@ -52,6 +52,11 @@ Properties	boolean Powered [readwrite]
 			off (and is available RF-Killed) while True means
 			that the technology is enabled.
 
+			When this property is changed while the Manager
+			OfflineMode property is true, ConnMan also remembers
+			whether the technology should be powered down the
+			next time offline mode is entered.
+
 		boolean Connected [readonly]
 
 			Boolean representing if a technology is connected.
--- a/connman/src/technology.c
+++ b/connman/src/technology.c
@@ -78,6 +78,7 @@ struct connman_technology {
 	char *tethering_passphrase;
 
 	bool enable_persistent; /* Save the tech state */
+	bool disable_in_offlinemode; /* Save if offline mode powers it off */
 
 	GSList *driver_list;
 
@@ -178,6 +179,74 @@ static const char *get_name(enum connman_service_type type)
 	return NULL;
 }
 
+static bool technology_default_disable_in_offlinemode(
+		enum connman_service_type type)
+{
+	switch (type) {
+	/*
+	 * Technologies that do not transmit radio signals should
+	 * stay on by default.
+	 */
+	case CONNMAN_SERVICE_TYPE_ETHERNET:
+	case CONNMAN_SERVICE_TYPE_GPS:
+	case CONNMAN_SERVICE_TYPE_GADGET:
+		return false;
+	/*
+	 * These types are not offline-mode-managed technologies.
+	 */
+	case CONNMAN_SERVICE_TYPE_UNKNOWN:
+	case CONNMAN_SERVICE_TYPE_SYSTEM:
+	case CONNMAN_SERVICE_TYPE_VPN:
+		return false;
+	/*
+	 * All transmitting radio technologies are powered down by default.
+	 */
+	case CONNMAN_SERVICE_TYPE_WIFI:
+	case CONNMAN_SERVICE_TYPE_BLUETOOTH:
+	case CONNMAN_SERVICE_TYPE_CELLULAR:
+	case CONNMAN_SERVICE_TYPE_P2P:
+		return true;
+	}
+
+	return true;
+}
+
+static bool technology_disable_in_offlinemode(
+		struct connman_technology *technology)
+{
+	return technology && technology->disable_in_offlinemode;
+}
+
+static bool technology_persistent_powered(
+		struct connman_technology *technology, bool offlinemode)
+{
+	if (!technology)
+		return false;
+
+	return technology->enable_persistent &&
+		(!offlinemode || !technology_disable_in_offlinemode(technology));
+}
+
+static void technology_set_persistent_powered(
+		struct connman_technology *technology, bool powered)
+{
+	if (!technology)
+		return;
+
+	if (global_offlinemode) {
+		technology->disable_in_offlinemode = !powered;
+
+		/*
+		 * Enabling a technology in offline mode should also make it
+		 * stay enabled when returning to normal mode.
+		 */
+		if (powered)
+			technology->enable_persistent = true;
+	} else {
+		technology->enable_persistent = powered;
+	}
+}
+
 static void technology_save(struct connman_technology *technology)
 {
 	GKeyFile *keyfile;
@@ -199,6 +268,9 @@ static void technology_save(struct connman_technology *technology)
 
 	g_key_file_set_boolean(keyfile, identifier, "Enable",
 				technology->enable_persistent);
+
+	g_key_file_set_boolean(keyfile, identifier, "DisableInOfflineMode",
+				technology->disable_in_offlinemode);
 
 	g_key_file_set_boolean(keyfile, identifier, "Tethering",
 				technology->tethering_persistent);
@@ -487,6 +559,18 @@ static int technology_load_values(struct connman_technology *technology,
 	}
 
 	enable = g_key_file_get_boolean(keyfile, identifier,
+					"DisableInOfflineMode", &error);
+	if (!error) {
+		technology->disable_in_offlinemode = enable;
+	} else {
+		technology->disable_in_offlinemode =
+			technology_default_disable_in_offlinemode(
+						technology->type);
+		need_saving = true;
+		g_clear_error(&error);
+	}
+
+	enable = g_key_file_get_boolean(keyfile, identifier,
 					"Tethering", &error);
 	if (!error) {
 		technology->tethering_persistent = enable;
@@ -523,6 +607,9 @@ static void technology_load(struct connman_technology *technology)
 			technology->enable_persistent = true;
 		else
 			technology->enable_persistent = false;
+		technology->disable_in_offlinemode =
+			technology_default_disable_in_offlinemode(
+						technology->type);
 		return;
 	}
 
@@ -863,7 +950,9 @@ static int technology_enabled(struct connman_technology *technology)
 		struct connman_technology *p2p;
 
 		p2p = technology_find(CONNMAN_SERVICE_TYPE_P2P);
-		if (p2p && !p2p->enabled && p2p->enable_persistent)
+		if (p2p && !p2p->enabled &&
+				technology_persistent_powered(p2p,
+						global_offlinemode))
 			technology_enabled(p2p);
 	}
 
@@ -904,7 +993,7 @@ static int technology_enable(struct connman_technology *technology)
 	if (technology->pending_reply)
 		return -EBUSY;
 
-	if (connman_setting_get_bool("PersistentTetheringMode")	&&
+	if (connman_setting_get_bool("PersistentTetheringMode") &&
 					technology->tethering)
 		set_tethering(technology, true);
 
@@ -1028,8 +1117,7 @@ static gboolean enable_delayed(gpointer user_data)
 	switch (err) {
 	case -EBUSY:
 		/* Make sure the pending reply does not block and continue */
-		if (technology_send_pending_reply(technology, -ECANCELED) ==
-					-ECOMM)
+		if (technology_send_pending_reply(technology, -ECANCELED) == -ECOMM)
 			connman_warn("could not reply to pending request");
 
 		return G_SOURCE_CONTINUE;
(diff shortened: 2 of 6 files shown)


connman/src/technology.c
```diff
@@ -2103,7 +2151,8 @@ bool __connman_technology_enable_from_config()
 			 * enable_persistent to make sure tech is enabled
 			 * when leaving offline mode.
 			*/
-			if (offlinemode) {
+			if (offlinemode &&
+				technology_disable_in_offlinemode(technology)) {
```
LaakkonenJussi: FYI: [coding style M4](https://github.com/sailfishos/connman/blob/master/connman/doc/coding-style.txt#L78 )

```

Example:
1)
if (call->status == CALL_STATUS_ACTIVE ||
	call->status == CALL_STATUS_HELD) {  // wrong
	connman_dbus_dict_append();

2)
if (call->status == CALL_STATUS_ACTIVE ||
		call->status == CALL_STATUS_HELD) {  // correct
	connman_dbus_dict_append();
```

connman/src/technology.c
```diff
@@ -2268,15 +2319,17 @@ int __connman_technology_add_rfkill(unsigned int index,
...
 	 * Depending on softblocked state we unblock/block according to
-	 * offlinemode and persistente state.
+	 * offlinemode policy and persistent state.
 	 */
 	if (technology->softblocked &&
-				!global_offlinemode &&
-				technology->enable_persistent)
+			(!global_offlinemode ||
+			 !technology_disable_in_offlinemode(technology)) &&
+			technology->enable_persistent)
```
LaakkonenJussi: Usually in these cases there are just tabs added, not aligned with spaces with anything above the line.

connman/src/technology.c
```diff
@@ -2268,15 +2319,17 @@ int __connman_technology_add_rfkill(unsigned int index,
...
 		return __connman_rfkill_block(type, false);
 	else if (!technology->softblocked &&
-		(global_offlinemode ||
-				!technology->enable_persistent)) {
+			((global_offlinemode &&
+			  technology_disable_in_offlinemode(technology)) ||
+			 !technology->enable_persistent)) {
```
LaakkonenJussi: Same here as comment R2327

connman/src/technology.c
```diff
@@ -1866,7 +1913,8 @@ int __connman_technology_set_offlinemode(bool offlinemode)
 		struct connman_technology *technology = list->data;
 
 		if (offlinemode) {
-			err = technology_disable(technology);
+			if (technology_disable_in_offlinemode(technology))
+				err = technology_disable(technology);
```
LaakkonenJussi: Not sure if leaving the value to `-EINVAL` is correct in this case as the `technology_disable()` will eventually call in device.c the `__connman_device_disable()` which, in case there is no way to disable, returns `-EOPNOTSUPP`. While not being supported the `switch()` later would just use the `default` action. Especially since this does not seem to require the saving of the offline mode with `connman_technology_save_offlinemode()`?

connman/src/technology.c
```diff
@@ -1739,7 +1785,8 @@ int __connman_technology_add_device(struct connman_device *device)
 	}
 
 	if (technology->enable_persistent &&
-					!global_offlinemode) {
+			(!global_offlinemode ||
+			 !technology_disable_in_offlinemode(technology))) {
```
LaakkonenJussi: Coding style [M4](https://github.com/sailfishos/connman/blob/master/connman/doc/coding-style.txt#L78)

connman/src/technology.c
```diff
@@ -178,6 +179,34 @@ static const char *get_name(enum connman_service_type type)
...
+static bool technology_default_disable_in_offlinemode(
+					enum connman_service_type type)
+{
+	switch (type) {
+	case CONNMAN_SERVICE_TYPE_ETHERNET:
+	case CONNMAN_SERVICE_TYPE_GPS:
+		return false;
+	case CONNMAN_SERVICE_TYPE_UNKNOWN:
+	case CONNMAN_SERVICE_TYPE_SYSTEM:
+	case CONNMAN_SERVICE_TYPE_VPN:
```
LaakkonenJussi, line 199: This types do not even have tech driver registered, so, yeah, cannot handle disable even :)

connman/doc/technology-api.txt
```diff
@@ -52,6 +52,11 @@ Properties	boolean Powered [readwrite]
 			off (and is available RF-Killed) while True means
 			that the technology is enabled.
 
+			When this property is changed while Manager
+			OfflineMode is true, ConnMan also remembers whether
+			the technology should be powered down the next time
+			offline mode is entered.
```
LaakkonenJussi: "Manager OfflineMode" or "OfflineMode in Manager/set via Manager" ?

Is it more clear to say "should be powered up the next time when offline mode is exited"?

LaakkonenJussi (changes requested):

connman/src/technology.c
abranson: I wasn't sure: is 'GADGET' the USB developer networking? Should I include that with GPS and ETHERNET as something that isn't really very relevant to flight mode?

LaakkonenJussi: Well since it is tethering it will be disabled with the transport anyways. So maybe no reason to do that.

connman/doc/technology-api.txt
abranson: That's not this property. This is for remembering if you want e.g. Bluetooth to remain on next time you enter flight mode, because you had it turned on last time you were in flight mode.

connman/src/technology.c
abranson: Unless they're tethering through ethernet?

LaakkonenJussi: I'm not sure if it will work. This is highly Sailfish OS specific thing and we do not have the possibility to have 2 USB adapters at the same time, other as developer connection to a laptop and other connecting to ethernet. I wonder how the tethering in such case would even work...

connman/doc/technology-api.txt
LaakkonenJussi: Ah, yeah, right. I think I read these before I went through other changes.

connman/src/technology.c
abranson: Or even sharing wifi over usb. I think because it's not possible in UI to turn it back on I'd rather have it default enabled in flight mode if that's ok with you. It won't hurt anyway.

connman/src/technology.c
abranson: Think this is ok now? This whole section is difficult to parse the tabs of tbh.

connman/src/technology.c
abranson: Right, that was all very confusing so I've replaced the rolling err with an accumulated technology_offlinemode_result_ok that should only flag the thing as failed if anything actually did return an error. I also found exiting flight mode failed a bit too easily, and reverted back into it which I don't think is ever a good idea.

LaakkonenJussi (commented):
Looks good now but maybe before merging I could do some testing next week with this.

LaakkonenJussi (approved):

connman/src/technology.c
```diff
@@ -2141,6 +2211,37 @@ bool __connman_technology_enable_from_config()
...
 						err ? strerror(-err) : "ok");
+		} else {
+			if (!technology->enabled) {
+				if (offlinemode &&
+						technology->enable_persistent &&
+						technology_disable_in_offlinemode(
+							technology)) {
+					DBG("tech %p/%s not enabled in offlinemode",
+						technology,
+						get_name(technology->type));
```
LaakkonenJussi: These can fit on the same line?

in M4 of coding-style.txt it is said:

```
If the line being wrapped is a function call or function declaration, the
preferred style is to indent at least past the opening parenthesis. Indenting
further is acceptable as well (as long as you don't hit the 80 character
limit).

If this is not possible due to hitting the 80 character limit, then indenting
as far as possible to the right without hitting the limit is preferred.

Example:

1)
gboolean sim_ust_is_available(unsigned char *service_ust, unsigned char len,
		enum sim_ust_service index); // worse

2)
gboolean sim_ust_is_available(unsigned char *service_ust, unsigned char len,
						enum sim_ust_service index);
						// better
```

The style is usually go past the parenthesis to ease up reading of the code.

connman/src/technology.c
```diff
@@ -2141,6 +2211,37 @@ bool __connman_technology_enable_from_config()
...
+				} else if (!technology->enable_persistent) {
+					DBG("tech %p/%s already disabled",
+						technology,
+						get_name(technology->type));
```
LaakkonenJussi: Same here as on R2222

connman/src/technology.c
```diff
@@ -2219,7 +2320,9 @@ static bool technology_apply_rfkill_change(struct connman_technology *technology
 			DBG("%s is switched on.", get_name(technology->type));
 			technology_dbus_register(technology);
 
-			if (global_offlinemode)
+			if (global_offlinemode &&
+					technology_disable_in_offlinemode(
+						technology))
```
LaakkonenJussi: Also same as in R2222

connman/src/technology.c
```diff
@@ -2066,7 +2167,8 @@ bool __connman_technology_enable_from_config()
 			DBG("technology %p/%s hardblocked, not set as %s",
 						technology,
 						get_name(technology->type),
-						technology->enable_persistent ?
+						technology_persistent_powered(technology,
+							offlinemode) ?
```
LaakkonenJussi: also here the same as with R2222

LaakkonenJussi (approved):
Seems to work ok. Only hiccup is with some user change scenarios but that might be how flight mode is handled in general when changing the user, especially if the modes are equal.

LGTM, with few cosmetic notes.

connman/src/technology.c
LaakkonenJussi: I guess this is ok. Lets change if something comes up, which I doubt.

connman/src/technology.c
```diff
@@ -2121,25 +2188,46 @@ bool __connman_technology_enable_from_config()
...
 			DBG("tech %p/%s disabled set as enabled, result %s",
-						technology,
-						get_name(technology->type),
+						technology, get_name(technology->type),
+						err ? strerror(-err) : "ok");
+		} else {
+			if (!technology->enabled) {
+				if (offlinemode &&
+					technology->enable_persistent &&
+					technology_disable_in_offlinemode(technology)) {
```
LaakkonenJussi: This and the lines above should be like
```suggestion
			if (offlinemode &&
						technology->enable_persistent &&
						technology_disable_in_offlinemode(technology)) {
					DBG("tech %p/%s not enabled in offlinemode",
```

connman/src/technology.c
```diff
@@ -2121,25 +2188,46 @@ bool __connman_technology_enable_from_config()
...
+					DBG("tech %p/%s not enabled in offlinemode",
+						technology, get_name(technology->type));
```
LaakkonenJussi: Maybe one more tab as in many others?

connman/src/technology.c
```diff
@@ -2121,25 +2188,46 @@ bool __connman_technology_enable_from_config()
...
+				} else if (!technology->enable_persistent) {
+					DBG("tech %p/%s already disabled",
+						technology, get_name(technology->type));
```
LaakkonenJussi: Maybe one more tab as in many others?

connman/src/technology.c
```diff
@@ -1141,9 +1229,9 @@ static DBusMessage *set_property(DBusConnection *conn,
 	if (dbus_message_iter_get_arg_type(&iter) != DBUS_TYPE_VARIANT)
 		return __connman_error_invalid_arguments(msg);
 
-	if (__connman_access_tech_set_property(get_tech_access_policy(),
-		name, dbus_message_get_sender(msg), CONNMAN_ACCESS_ALLOW) !=
-						CONNMAN_ACCESS_ALLOW) {
+	if (__connman_access_tech_set_property(get_tech_access_policy(), name,
+				dbus_message_get_sender(msg), CONNMAN_ACCESS_ALLOW) !=
```
LaakkonenJussi: Extra space here at the end

connman/src/technology.c
```diff
@@ -1141,9 +1229,9 @@ static DBusMessage *set_property(DBusConnection *conn,
 	if (dbus_message_iter_get_arg_type(&iter) != DBUS_TYPE_VARIANT)
 		return __connman_error_invalid_arguments(msg);
 
-	if (__connman_access_tech_set_property(get_tech_access_policy(),
-		name, dbus_message_get_sender(msg), CONNMAN_ACCESS_ALLOW) !=
-						CONNMAN_ACCESS_ALLOW) {
+	if (__connman_access_tech_set_property(get_tech_access_policy(), name,
+				dbus_message_get_sender(msg), CONNMAN_ACCESS_ALLOW) != 
+			CONNMAN_ACCESS_ALLOW) {
```
LaakkonenJussi, line 1234: Few more tabs?

connman/src/technology.c
abranson: I don't understand this one sorry.

PR #128, opened 2026-04-20
Wifi WPA3 support level via D-BUS

LaakkonenJussi:


direc85 (approved):
Seems to work on my devices; one supports WPA3 and another one doesn't.

PR #132, opened 2026-04-29
Add Supported property, set unsupported WiFi's favorite and autoconnect false

LaakkonenJussi:
Add a new property to tell if the network is supported on the device or not. The service can be available, i.e., show the signal strength but If the WiFi service cannot be supported due to lack of WPA3 support make the service not-autoconnectable and disable favorite option.

### DIFF
--- a/connman/src/service.c
+++ b/connman/src/service.c
@@ -391,6 +391,7 @@ struct connman_service {
 	bool removing;
 	char *wpa3_sae_pwe;
 	bool wpa3_sae_check_mfp;
+	bool supported;
 };
 
 static const char *service_get_access(struct connman_service *service);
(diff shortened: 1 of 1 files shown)


direc85 (approved):

LaakkonenJussi:
I now think that we need to make the `supported` a boolean available over D-Bus as read-only as a service property. Having `supported` tied to `available` means that in connection selector the network does not show up hence being not available.

connman/src/service.c
```diff
@@ -2504,6 +2531,52 @@ static void service_set_removing(struct connman_service *service,
...
+		DBG("%s update autoconnect and favorite", service->identifier);
+		connman_service_set_autoconnect(service, false);
+		__connman_service_set_favorite(service, false);
+	}
+
+	if (service->supported != was_supported) {
+		DBG("%s supported changed to %s", service->identifier,
+					service->supported ? "true" : "false");
+		service_boolean_changed(service, &service_supported);
+	} else {
```
LaakkonenJussi: TODO: remove this, unnecessary DBG()'s. Maybe some above could be dropped.

direc85 (approved):

PR #133, opened 2026-05-04
Update MaxRate D-Bus property when changed, get value from kernel

LaakkonenJussi:
Drop the maxrate use via libgsupplicant and use the kernel value directly. Update the value periodically. This requires network.c to update the value as well.

Added libnl and libnl-genl (same package) requirement for build.

### DIFF
--- a/connman/plugins/sailfish_wifi.c
+++ b/connman/plugins/sailfish_wifi.c
@@ -45,6 +45,11 @@
 #include <sys/socket.h>
 #include <linux/if.h>
 
+#include <netlink/genl/genl.h>
+#include <netlink/genl/ctrl.h>
+#include <netlink/netlink.h>
+#include <linux/nl80211.h>
+
 #define WIFI_SERVICE_PREFIX "wifi_"
 #define NETWORK_BGSCAN "simple:30:-65:300"
 #define WIFI_BSSID_LEN 6
@@ -109,6 +114,9 @@
 #define NETWORK_EAP_DEFAULT                     "default"
 
 #define WMTWIFI_PATH				"/dev/wmtWifi"
+
+#define NL80211					"nl80211"
+#define NL80211_MAXRATE_TIMEOUT			1		/* Seconds */
 
 enum supplicant_events {
 	SUPPLICANT_EVENT_VALID,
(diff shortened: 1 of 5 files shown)


connman/plugins/sailfish_wifi.c
```diff
@@ -807,6 +818,281 @@ static gboolean wifi_network_connecting(struct wifi_network *net)
...
+		return NL_SKIP;
+	}
+
+	if (stainfo[NL80211_STA_INFO_TX_BITRATE]) {
+		tx = parse_bitrate(stainfo[NL80211_STA_INFO_TX_BITRATE]);
+		DBG("TX bitrate: %.1f Mbps", tx / 10.0);
+
+		/*
+		 * Use tx rate to inform upper layers. This is the common way.
+		 * Values are reported as kbps units -> scale to bytes/s
```
mlehtima: Shouldn't the units be 100 kbps?

connman/plugins/sailfish_wifi.c
```diff
@@ -807,6 +818,281 @@ static gboolean wifi_network_connecting(struct wifi_network *net)
...
+					genlmsg_attrlen(gnlh, 0), NULL)) {
+		DBG("failed to parse attribute stream: %d/%s", errno,
+						strerror(errno));
+		return NL_SKIP;
+	}
+
+	if (!attrs[NL80211_ATTR_STA_INFO])
+		return NL_SKIP;
+
+	if (nla_parse_nested(stainfo,NL80211_STA_INFO_MAX,
```
mlehtima: Missing space after ,

connman/plugins/sailfish_wifi.c
LaakkonenJussi: Yeah they are. I think I forgot it from this comment.

LaakkonenJussi: Added.

connman/plugins/sailfish_wifi.c
LaakkonenJussi: Added space.

mlehtima (approved):

PR #134, opened 2026-05-12
Upstream sync from nlnet work

LaakkonenJussi:
Device, settings and systemd fixes from NLnet work. Some things noticed missing when making the upstream patches.

### DIFF
--- a/connman/src/connman.h
+++ b/connman/src/connman.h
@@ -712,6 +712,7 @@ void __connman_device_cleanup(void);
 
 void __connman_device_list(DBusMessageIter *iter, void *user_data);
 
+enum connman_device_type __connman_device_string2type(const char *str);
 enum connman_service_type __connman_device_get_service_type(struct connman_device *device);
 struct connman_device *__connman_device_find_device(enum connman_service_type type);
 int __connman_device_request_scan(enum connman_service_type type);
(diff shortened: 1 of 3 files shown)


connman/vpn/vpn-provider.c
```diff
@@ -3016,11 +3016,14 @@ static int set_string(struct vpn_provider *provider,
 		hide_value ? "<not printed>" : value);
 
 	if (g_str_equal(key, "Type")) {
-		if (!g_strcmp0(provider->type, value))
+		char *type = g_ascii_strdown(value, -1);
```
spiiroin: can it be assumed that value is never null?
(other functions used here, like g_free, g_strcmp0, g_strdup explicitly allow null args but g_ascii_strdown does not)

spiiroin (approved):

connman/vpn/vpn-provider.c
LaakkonenJussi: Well, yeah, it can be, but that was actually a problem also before the change. If value is `NULL` it does not match to `provider->type` and `g_ascii_strdown()` would have been called with `NULL`. Adding a check would not hurt, though.

LaakkonenJussi: Fixed, allow to use `NULL` to clear the value as before.

PR #135, opened 2026-05-13
Support wmtWifi type device tethering on Sailfish OS WiFi plugin

LaakkonenJussi:
Add pre/post operation support when enabling tethering. This triggers, with -EINPROGRESS error a delayed start of tethering so the newly brought up interfaces can be added and the plugin has time to react to them.

A two new options are added to settings in order to specify the wmtWifi device sequence to be sent. 

Sailfish OS WiFi plugin records the newly created AP interface index and uses that device when tethering is pending for further actions.

### DIFF
--- a/connman/plugins/sailfish_wifi.c
+++ b/connman/plugins/sailfish_wifi.c
@@ -106,8 +106,9 @@
 #define NETWORK_KEY_WIFI_SAE_PWE                "WiFi.SAEPWE"
 #define NETWORK_KEY_WIFI_SAE_CHECK_MFP          "WiFi.SAECheckMFP"
 
-
 #define NETWORK_EAP_DEFAULT                     "default"
+
+#define WMTWIFI_PATH				"/dev/wmtWifi"
 
 enum supplicant_events {
 	SUPPLICANT_EVENT_VALID,
@@ -322,6 +323,9 @@ struct wifi_plugin {
 	gulong supplicant_event_id[SUPPLICANT_EVENT_COUNT];
 	struct connman_technology *tech;
 	gboolean running;
+	gboolean wmtWiFi;
+	gboolean tethering_pending;
+	int tethering_index;
 	GSList *devices;
 };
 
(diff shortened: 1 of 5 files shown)


mkosola:
There is "support support" in commit message.

connman/plugins/sailfish_wifi.c
```diff
@@ -4729,7 +4732,27 @@ static int wifi_plugin_set_tethering(struct wifi_plugin *plugin,
...
+					ifname = connman_inet_ifname(dev->ifi);
+					DBG("Found new AP interface %d:%s",
+							dev->ifi, ifname);
+					plugin->tethering_pending = false;
+					plugin->tethering_index = 0;
+					ap_dev = dev;
+					g_free(ifname);
+				} else {
+					ap_dev = dev;
+				}
```
direc85: Since one of these assignments is taken either way, it could be after the if-else block.

LaakkonenJussi:
> There is "support support" in commit message.

Heh, so it seems. Thanks for pointing it out. Fixed.

direc85 (approved):
Seems to work. One minor comment, fine either way!

connman/plugins/sailfish_wifi.c
LaakkonenJussi: Yeah, like 39ed64d9c7fdbf727993953a394c12a66ae87e39 ? Not sure if we should select a device for `wmtWifi` if we're not pending... But that should not happen. Now I wonder if it is a possibility in some scenario?

LaakkonenJussi: Seems to work on wmtWifi device and normal device. Marking as resolved.

piggz:
Should this
Obsoletes: sailfish-connman-plugin-tethering-wmtwifi
?

LaakkonenJussi:
> Should this Obsoletes: sailfish-connman-plugin-tethering-wmtwifi ?

Yes, that plugin is garbage anyways. It should not be attempted to be handled like that.

This has just been tested on one wmtWiFi device, so when confirmed to be working (the sequences need to be defined for each device in a .conf file at main.conf.d/) we can add that obsoletes. That could be added before the release this will be part of.

PR #137, opened 2026-05-26
Add a wmtWifi dual mode property to D-Bus and use it in WLAN plugin

LaakkonenJussi:
Add a WifiWMTDualMode boolean configuration option and expose it as a read only D-Bus property. This can be used to define the Sailfish OS WiFi plugin behavior with wmtWifi devices when tethering. Value for the option is by default `false` that defines dual mode not being used and when tethering is enabled the normal WLAN device is disabled in favor of using the AP device. In dual mode (`true`) WLAN device remains active and is not disabled/enabled when AP mode is activated.

The default behavior is, thus, identical to what it is without wmtWifi device. This does not require any UI changes but allows to enable the dual mode in configuration when making the UI changes.

### DIFF
--- a/connman/plugins/sailfish_wifi.c
+++ b/connman/plugins/sailfish_wifi.c
@@ -5325,6 +5325,13 @@ static int wifi_tech_driver_set_tethering_prepare(
 	if (!is_wmtWiFi(tech))
 		return -ENODEV;
 
+	/*
+	 * Prevent to send multiple enable sequences. Allow to always send
+	 * disable sequence.
+	 */
+	if (enabled && wifi_plugin->tethering_pending)
+		return -EALREADY;
+
 	err = send_wmtWifi_sequence(tech, enabled);
 	if (err) {
 		connman_error("Failed to prepare for tethering: %d/%s", err,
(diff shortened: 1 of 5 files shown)


connman/plugins/sailfish_wifi.c
```diff
@@ -4708,6 +4710,31 @@ static struct connman_device_driver wifi_device_driver = {
...
+						struct wifi_device *exclude)
+{
+	GSList *l;
+	for (l = plugin->devices; l; l = l->next) {
+		struct wifi_device *dev = l->data;
+		if (dev == exclude || dev->tethering)
+			continue;
+
+		if (enable) {
+			if (dev->enable_after_tethering && !dev->tethering) {
```
mlehtima: Isn't the !dev->tethering useless because above we continue if dev->tethering is true?

LaakkonenJussi: Yeah it seems to be.

LaakkonenJussi: Dropped that check.

mlehtima (approved):

PR #138, opened 2026-05-29
Fix DNS leak with VPN and two connected devices when using systemd-resolved

LaakkonenJussi:
The scenario this addresses is when:
1) Device A (e.g., mobile data) is connected and connect VPN over it
2) Connect a WLAN connection (or any that is sorted above Device A)
3) Let VPN auto-connect over WLAN
4) Disable WLAN and let Device A to connect and wait VPN to auto-connect
5) DNSs severs of Device A remain active

This can be observed with, for example:

```
pattern=$(IFS='|'; echo "$*")
watch -n 1 "systemd-resolve --status|awk '/$pattern/{f=1} f; /^$/{f=0}'"
```

By giving all interface names as arguments.

The change here first checks if the interface index is being used as a default to be able to update the DNS servers. Updates to DNS servers must be also allowed before the device is being even enabled, so for this reason the update is allowed without default service being set as the service might be getting enabled later. This change addresses the both cases, especially the one described above with VPNs that are used as default route, i.e., getting set as default service.

### DIFF
--- a/connman/src/dns-systemd-resolved.c
+++ b/connman/src/dns-systemd-resolved.c
@@ -226,12 +226,75 @@ static bool is_empty(struct dns_interface *iface)
 	return (!iface->domains && !iface->servers);
 }
 
+/*
+ * If the default service exists and it is different do not allow to update this
+ * service. Only the default service's DNS should be used. The services may be
+ * updated after the service was disabled but that would then block use of the
+ * service at startup as DNSs and domains are updated before the service is
+ * enabled. This makes it possible to handle both cases, when the service is
+ * just initializing or when a service gets updates while not being as default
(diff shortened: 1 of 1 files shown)


mlehtima (approved):

mlehtima (approved):

LaakkonenJussi:
This needs some minor fix still. When a VPN is used as a non-default route the DNS servers of it should be kept, so the resources behind the VPN can be accessed.

LaakkonenJussi:
> This needs some minor fix still. When a VPN is used as a non-default route the DNS servers of it should be kept, so the resources behind the VPN can be accessed.

I think it is fixed now.

mlehtima (approved):

PR #142, opened 2026-07-22
Change setting to hash table approach, add WiFi options tests

LaakkonenJussi:
Use the setting hash table approach made in the NLnet project ([original PR](https://github.com/LaakkonenJussi/connman_ng/pull/4), and [updated PR](https://github.com/LaakkonenJussi/connman_ng/pull/5)). This changes the process but not the outcome for processing the settings. New config options are easier and more straightforward to add.

Added unit tests for the new WiFi options and improved as well as fixed the unit test for setting a bit.

Included small string2type fixes for device and service as well.


connman/src/setting.c
```diff
@@ -77,517 +99,120 @@ static char *default_blacklist[] = {
...
-					CONF_TYPE_BOOL},
-	{ 0 }
+/* Union for storing the values */
+union config_value {
+	bool bool_val;
+	unsigned int uint_val;
+	double double_val;
+	char *str_val;
+	char **str_array_val;
+	unsigned int *int_array_val;
```
spiiroin: unsigned int value = `uint_val` vs array of unsigned ints = `int_array_val` - maybe it should be `uint_array_val` ?

connman/src/setting.c
```diff
@@ -724,365 +383,1155 @@ static gboolean check_wpa3_sae_pwe(const char *str)
...
+	if (!opt)
+		return NULL;
+
+	if (opt->opt_type != CONF_TYPE_STR &&
+					opt->opt_return_type != CONF_TYPE_STR)
+		return NULL;
+
+	if (opt->current_val.str_val) {
+		return opt->current_val.str_val;
+	} else {
```
spiiroin: nit: superfluous `else`

connman/src/setting.c
```diff
@@ -724,365 +383,1155 @@ static gboolean check_wpa3_sae_pwe(const char *str)
...
+	g_strfreev(opt->current_val.str_array_val);
+	opt->current_val.str_array_val = g_strdupv(value);
+}
 
-	for (i = 0; i < len; i++)
-		cb(value[i]);
+static void set_int_array_value(struct config_option *opt, unsigned int *value)
+{
+	if (opt->opt_type != CONF_TYPE_UINTARR)
+		return;
```
spiiroin, line 553: set_str_array_value() is clean: on success it stores duplicated `value` and caller always retains ownership of the original.

whereas this set_int_array_value() looks like bug / memory leak waiting to happen: on success `value` ownership is taken from caller, on failure path not - and caller can't know which way it went.

atm used only from one place, in a way that avoids this problem. still, a bit iffy / fragile.

connman/src/setting.c
```diff
@@ -724,365 +383,1155 @@ static gboolean check_wpa3_sae_pwe(const char *str)
...
+	if (opt->opt_type != CONF_TYPE_HASHTABLE)
+		return;
 
-	if (*ptr)
-		g_free(*ptr);
+	if (opt->current_val.hash_table_val)
+		g_hash_table_destroy(opt->current_val.hash_table_val);
 
-	*ptr = value;
+	opt->current_val.hash_table_val = value;
```
spiiroin: ditto, reference is taken over if we get here, not if we have hit the early return...
could increment reference here, or decrement on early return depending what ought to happen

connman/src/setting.c
```diff
@@ -724,365 +383,1155 @@ static gboolean check_wpa3_sae_pwe(const char *str)
...
+				CONF_ONLINE_CHECK_SUCCESSES_THRESHOLD));
+}*/
+
+static struct config_option config_options[] = {
+	/* BackgroundScanning */
+	{
+		.opt_key = CONF_BG_SCAN,
+		.opt_type = CONF_TYPE_BOOL,
+		.opt_return_type = CONF_TYPE_BOOL,
+		.default_val.bool_val = true,
```
spiiroin, line 790: nit: `current_val` is not explicitly initialized in any of these. IIRC default union initialization happens per the first listed member. Which has type bool -> strictly speaking values that are booleans are initialized ok (to false), others to "something".

One way to deal with this is to have dummy array as the first member, but given that this goes to bss and is not taken from stack, probably ok as long as null pointers, zero value doubles, etc are of  "all bits zero" type.

Alternatively, initialize_default_values() looks like something that could do this correclty / type?

spiiroin (approved):
Some remarks/suggestions, no clear / real errors spotted.

connman/src/setting.c
LaakkonenJussi: Yeah, better that way. Changed in 5517dda4757c8d775754cab2890883d7c2d21525

connman/src/setting.c
LaakkonenJussi: Yes. Dropped.

connman/src/setting.c
LaakkonenJussi: Changed this to duplicate the value always, also with hash table. Change in 5517dda4757c8d775754cab2890883d7c2d21525

connman/src/setting.c
LaakkonenJussi: Yeap, added a `ref()` call with 5517dda4757c8d775754cab2890883d7c2d21525

connman/src/setting.c
LaakkonenJussi: You mean something like this done with 5517dda4757c8d775754cab2890883d7c2d21525 ? Should not the default_val copy handle this for regular types, and now the current value for str/uint array and hash table is set to `NULL` explicitly, if not in the already set default value setters.

spiiroin: My idea was roughly: add a switch(type)  that sets .current_va.xxx = sane default for type before the current default processing / copying default value to current that already is there. so that int/double/ptr field in union gets a zero/null value even if default initialization zeroes only sizeof(bool) portion of it. Or alternatively add char dummy_val[sizeof(double)]; before bool bool_val (with language lawyer caveat for doubles and pointers). However, as this is statically initialized block, not sure if it is worthwhile - it is going to be all zeroes anyway.

LaakkonenJussi: Ok, well I think we can live with this, hence the static initialization.

PR #143, opened 2026-07-29
Add VPN IPv4 and IPv6 addresses separately, fix WG AllowIPs routes, fix CIDR conversion

LaakkonenJussi:
In some cases having both in the same D-Bus message one can override the other as there is only one connman_ipaddress struct. Add separate structs to avoid any problems.

Fix WireGuard AllowIPs routes, they need to be added without the endpoint as gateway. Allow this in vpn-provider.c and fix the default route handling. The default route must not be dropped by WireGuard plugin, but the vpn.c needs to handle that there are no doubles. Thus, search the default route with the specified IP family from the hash table before adding instead of relying a boolean set elsewhere.

Bit shifting by 32 on 32 bit integers is undefined behavior in C. Fix this by using `0` netmask in `cidr_to_netmask()` conversion.


connman/vpn/plugins/wireguard.c
```diff
@@ -349,13 +349,21 @@ struct wg_ipaddresses {
...
+
 		return g_strdup_printf("%d.%d.%d.%d",
-				((0xffffffff << (32 - cidr)) >> 24) & 0xff,
-				((0xffffffff << (32 - cidr)) >> 16) & 0xff,
-				((0xffffffff << (32 - cidr)) >> 8) & 0xff,
-				((0xffffffff << (32 - cidr)) >> 0) & 0xff);
+				(mask >> 24) & 0xff,
+				(mask >> 16) & 0xff,
+				(mask >> 8) & 0xff,
+				(mask >> 0) & 0xff);
```
spiiroin, line 366: nit: '%d' formatting vs unsigned / uint32_t values

spiiroin (approved):

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Yeah, `%u` works better.

PR #144, opened 2026-08-12
Very WIP: Jb62307

LaakkonenJussi:


connman/vpn/plugins/wireguard.c
```diff
@@ -632,130 +715,381 @@ static void resolve_endpoint_cb(GResolvResultStatus status,
...
+{
+	const char *option;
+	const char *endpoint;
+	int family;
+	int failedPeers = 0;
+	int err = 0;
+	int i;
+
+	for (i = 0; i < peercount; i++) {
+		struct wg_peer *peer = info->peer;
```
abranson: Should this be before the loop? It gets set to the first peer on each iteration.

connman/vpn/plugins/wireguard.c
abranson, line 207: This function is now getting called multiple times, and this would erase the previous result.

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Yeah, changed to be initialized in the `for()`.

connman/vpn/plugins/wireguard.c
LaakkonenJussi: True. Changed to set this to false when one default route is detected. It should not return to true when next peer(s) are not having default route. In fact, only one should have default route set if I'm understood correctly.

connman/vpn/plugins/wireguard.c
```diff
@@ -507,47 +516,68 @@ static bool sockaddr_cmp_addr(struct sockaddr_u *a, struct sockaddr_u *b)
...
-	if (info->remove_resolv_id)
-		g_source_remove(info->remove_resolv_id);
+	if (resolv->remove_resolv_id)
+		g_source_remove(resolv->remove_resolv_id);
 
-	if (info->resolv && info->resolv_id) {
+	if (resolv->resolv && resolv->resolv_id) {
 		DBG("cancel resolv lookup");
-		vpn_util_cancel_resolve(info->resolv, info->resolv_id);
+		vpn_util_cancel_resolve(resolv->resolv, resolv->resolv_id);
```
spiiroin, line 656: any reason not to do this 1st, before g_source_remove? now whatever is done, is done based on / using stale source id number -> looks iffy.

connman/vpn/plugins/wireguard.c
```diff
@@ -642,114 +716,395 @@ static void resolve_endpoint_cb(GResolvResultStatus status,
...
+
+			/* Add data to next and set it as last */
+			peer = peer->next_peer;
+			peer->flags = WGPEER_HAS_PUBLIC_KEY |
+						WGPEER_REPLACE_ALLOWEDIPS;
+			info->device.last_peer = peer;
+		/* Previous one failed, reuse the peer */
+		} else {
+			failedPeers++;
+			DBG("using skipped peer, failed: %d", failedPeers);
```
spiiroin: what is the point here?
this is all very confusing, but afaict a) given that this starts iterating from info->peer, the code seems to be prepared to see an already populated peer list b) if  such prepopulated list has more than one entry -> we trash the 1st entry over and over again using diffent `i` values or stte  c) in case of starting from an empty list, we are not going to end up here

connman/vpn/plugins/wireguard.c
```diff
@@ -802,84 +1156,43 @@ static int wg_connect(struct vpn_provider *provider,
...
-		err = parse_key(option, info->peer.preshared_key);
-		if (err) {
-			DBG("Failed to parse pre-shared key");
-			goto error;
-		}
+		err = create_singlepeer(info, &do_split_routing, &gateway4,
+					&gateway6);
+	} else {
+		char *end;
+		int peercount = g_ascii_strtoull(option, &end, 10);
```
spiiroin: if `end` is not used, might as well pass null?

applies for other places that use g_ascii_strtoull() / similar too

connman/vpn/plugins/wireguard.c
```diff
@@ -931,10 +1244,11 @@ static int wg_connect(struct vpn_provider *provider,
...
 	if (!err) {
 		/* Run DNS reresolve only for hostnames that require resolve. */
-		family = connman_inet_check_ipaddress(info->endpoint_fqdn);
-		if (family != AF_INET && family != AF_INET6) {
-			DBG("start DNS reresolve for %s", info->endpoint_fqdn);
-			run_dns_reresolve(info);
+		struct wg_peer_resolv *r;
+		for (r = info->resolv; r; r = r->next) {
+			r->data.info = info;
+			r->data.resolv = r;
```
spiiroin: nit: other loops use `resolv` variable for iteration

connman/vpn/plugins/wireguard.c
```diff
@@ -642,114 +716,395 @@ static void resolve_endpoint_cb(GResolvResultStatus status,
...
+		if (family != AF_INET && family != AF_INET6) {
+			DBG("start DNS reresolve for %s", endpoint);
+			resolv = info->resolv;
+
+			if (!resolv) {
+				info->resolv = g_try_new0(struct wg_peer_resolv,
+								1);
+				resolv = info->resolv;
+			} else {
+				resolv->next = g_try_new0(struct wg_peer_resolv,
```
spiiroin: eh? if we have list head, leak and override already existing 2nd element. or am I missing something?

connman/vpn/plugins/wireguard.c
```diff
@@ -642,114 +716,395 @@ static void resolve_endpoint_cb(GResolvResultStatus status,
...
+
+			if (!resolv) {
+				err = -ENOMEM;
+				DBG("failed to allocate new resolv for #%d", i);
+				return -ENOMEM;
+			}
+
+			resolv->endpoint_fqdn = g_strdup(endpoint);
+			resolv->port = g_strdup(option);
+			resolv->id = i;
```
spiiroin: also, if the resolv list really gets truncated to max 2 entries, resulting `resolv->id` here is not really an index within this list / peer vs resolv lists do not match-> does it affect e.g. get_nth_peer(resolv->id) usage somehow?

connman/vpn/plugins/wireguard.c
spiiroin: hmm, I misread id names -> forget about it ;-)

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Renamed.

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Reworked this.

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Reworked, this was not fully thought out yet it seems.

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Yeah at least dropped in this section. Lets see others later, those are from upstream implementation anyways.

connman/vpn/plugins/wireguard.c
LaakkonenJussi: Just an idea to have the checks done and if they fail and peer has invalid content do not unnecessarily allocate a new one but to reuse the already allocated. Reworked this part too, looks now cleaner to understand maybe.

connman/vpn/plugins/wireguard.c
```diff
@@ -78,44 +73,106 @@ struct sockaddr_u {
...
 
+static char* get_wg_opt(const char *opt, bool peer_option, int index)
+{
+	if (!opt)
+		return NULL;
+
+	/* Keep backwards compatibility when single peer mode is in use. */
+	if (peer_option && index >= 0)
+		return g_strdup_printf("WireGuard.Peer%d.%s", index, opt);
+	else
```
spiiroin: nit: superfluous 'else'

connman/vpn/plugins/wireguard.c
```diff
@@ -78,44 +73,106 @@ struct sockaddr_u {
...
+
+		DBG("free resolv for peer #%d", tmp_resolv->id);
+		g_free(tmp_resolv->endpoint_fqdn);
+		g_free(tmp_resolv->port);
+		g_free(tmp_resolv);
+	}
+
+	peer = info->peer;
+	while (peer) {
+		struct wg_peer *tmp_peer = peer;
```
spiiroin, line 160: nit: like `temp_peer`,  `allowedip` is used only here, but is declared in function scope?

side note: all this custom next stuff - isn't there some list container thingamajick that would be used in connman?

connman/vpn/plugins/wireguard.c
```diff
@@ -642,114 +767,406 @@ static void resolve_endpoint_cb(GResolvResultStatus status,
...
+			switch (allowedip->family) {
+			case AF_INET:
+				if (!inet_ntop(allowedip->family,
+							&allowedip->ip4, addr,
+							INET6_ADDRSTRLEN)) {
+					DBG("ignore invalid IPv4 address");
+					continue;
+				}
+
+				break;
```
spiiroin, line 831: nit: empty lines before 'break' - is that intentional?

connman/vpn/plugins/wireguard.c
```diff
@@ -1000,14 +1382,16 @@ static int disconnect(struct vpn_provider *provider, int err)
...
-		g_source_remove(info->reresolve_id);
-
 	if (info->route_setup_id)
 		g_source_remove(info->route_setup_id);
 
-	if (info->resolv || info->resolv_id)
-		remove_resolv(info);
+	for (resolv = info->resolv; resolv; resolv = resolv->next) {
+		if (resolv->reresolve_id)
+			g_source_remove(resolv->reresolve_id);
```
spiiroin, line 1467: should the id value be cleared?

spiiroin (approved):
Some random cosmetic remarks

connman/vpn/plugins/wireguard.c
LaakkonenJussi: There is the foreach for the wireguard structs. Lets see if they fit here.

LaakkonenJussi: The `for_each` are just wrappers for... `for()` and when freeing the content I think current approach is safer, to take a tmp ptr and propagate to next before freeing.

